PR #9, opened 2019-04-22
sha2.c: Fix for systems that do not support non word alignment

mrdeep1:
dtls_sha256_transform() fails for esp-idf systems as the data parameter
is a uint32 and may not be aligned aligned to a word boundary when
processing the server-hello handshake.

sha2/sha2.c:

Make the data parameter byte aligned (as does most other sha256
implementations) for function dtls_sha256_transform().

### DIFF
--- a/sha2/sha2.c
+++ b/sha2/sha2.c
@@ -149,23 +149,69 @@ typedef u_int64_t sha2_word64;	/* Exactly 8 bytes */
 #define DTLS_SHA384_SHORT_BLOCK_LENGTH	(DTLS_SHA384_BLOCK_LENGTH - 16)
 #define DTLS_SHA512_SHORT_BLOCK_LENGTH	(DTLS_SHA512_BLOCK_LENGTH - 16)
 
+/*** ENDIAN / WORD ALIGNING Handling Functions ***/
 
-/*** ENDIAN REVERSAL MACROS *******************************************/
+static inline sha2_word32 get32be(const sha2_byte* data)
+{
 #if BYTE_ORDER == LITTLE_ENDIAN
-#define REVERSE32(w,x)	{ \
-	sha2_word32 tmp = (w); \
-	tmp = (tmp >> 16) | (tmp << 16); \
-	(x) = ((tmp & 0xff00ff00UL) >> 8) | ((tmp & 0x00ff00ffUL) << 8); \
+	sha2_word32 tmp = ((sha2_word32)data[0] << 24) |
+			  ((sha2_word32)data[1] << 16) |
+			  ((sha2_word32)data[2] <<  8) |
+			   (sha2_word32)data[3];
+	return tmp;
+#else /* BYTE_ORDER != LITTLE_ENDIAN */
+	sha2_word32 tmp;
+	MEMCPY_BCOPY(&tmp, data, sizeof(tmp));
+	return tmp;
+#endif /* BYTE_ORDER != LITTLE_ENDIAN */
 }
-#define REVERSE64(w,x)	{ \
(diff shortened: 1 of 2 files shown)


sha2/sha2.c
```diff
@@ -470,7 +471,7 @@ void dtls_sha256_transform(dtls_sha256_ctx* context, const sha2_word32* data) {
 
 #else /* SHA2_UNROLL_TRANSFORM */
 
-void dtls_sha256_transform(dtls_sha256_ctx* context, const sha2_word32* data) {
+void dtls_sha256_transform(dtls_sha256_ctx* context, const sha2_byte* data) {
```
obgm: The change from `sha2_word32` to `sha2_byte` (i.e. from `uint32_t` to `uint8_t`) breaks the `data++` if `BYTE_ORDER != LITTLE_ENDIAN`.

mrdeep1: Missed that in my testing as I was not running with BIG_ENDIAN code.  However, I suspect that the same issue (non aligned word access giving unexpected results) will be present on BIG ENDIAN code as well.  Looks like a lot of code re-writing needed.

[Finding the issue was an interesting time!]

obgm: Indeed.

The big endian code at least has not the issue with the REVERSE macros as they are no-ops except for the pointer increment. IMO your fix for `dtls_sha256_transform()` is a good start and just needs a small adjustment for the big endian case. (Also note that the type cast of `context->buffer` at the very beginning is a potential pitfall as it works on the assumption that the buffer always starts at a word boundary.)

**But** there are more alignment issues in the unrolled version of `dtls_sha265_transform()`. Uh!

mrdeep1: I have not fixed the W256 and W512 typecast for `context->buffer` as I think it very unlikely that they would not be the appropriate word aligned and changing the code would be complex to convert to byte aligned.

obgm:
For `-DSHA2_UNROLL_TRANSFORM`, the following needs to be fixed:

```
diff --git a/sha2/sha2.c b/sha2/sha2.c
index 4cdd668..cb6d90f 100644
--- a/sha2/sha2.c
+++ b/sha2/sha2.c
@@ -735,8 +735,8 @@ void dtls_sha512_init(dtls_sha512_ctx* context) {
 /* Unrolled SHA-512 round macros: */
 
 #define ROUND512_0_TO_15(a,b,c,d,e,f,g,h)      \
-       W512[j] = get64be(data);
-       data += 8;
+       W512[j] = get64be(data); \
+       data += 8; \
        T1 = (h) + Sigma1_512(e) + Ch((e), (f), (g)) + \
              K512[j] + W512[j]; \
        (d) += T1, \
@@ -754,7 +754,7 @@ void dtls_sha512_init(dtls_sha512_ctx* context) {
        (h) = T1 + Sigma0_512(a) + Maj((a), (b), (c)); \
        j++
 
-void dtls_sha512_transform(dtls_sha512_ctx* context, const sha2_word64* data) {
+void dtls_sha512_transform(dtls_sha512_ctx* context, const sha2_byte* data) {
        sha2_word64     a, b, c, d, e, f, g, h, s0, s1;
        sha2_word64     T1, *W512 = (sha2_word64*)context->buffer;
        int             j;
```

obgm:
And can you please also fix the duplicate word in the commit message?

mrdeep1:
Changes uploaded.

obgm:
Merge as all tests succeed with and without `-DSHA2_UNROLL_TRANSFORM` for all three hash functions on little-endian platform. (Code for big-endian looks correct to me, tests to be done.)

PR #10, opened 2019-04-23
hmac.h: Add in ESP-IDF sha256 versions

mrdeep1:
For ESP-IDF, use the ESP-IDF sha356 hash implementation to save on code
size.

### DIFF
--- a/hmac.h
+++ b/hmac.h
@@ -36,7 +36,19 @@ typedef sha256_context_t dtls_sha256_ctx;
 #define dtls_sha256_update(Ctx,Input,Len) sha256_update((Ctx), (Input), (Len))
 #define dtls_sha256_final(Buf,Ctx)        sha256_final((Ctx), (Buf))
 
-#else /* RIOT_VERSION */
+#elif defined ESP_PLATFORM && defined CONFIG_LIBSODIUM_USE_MBEDTLS_SHA
+#include "sodium/crypto_hash_sha256.h"
+
+typedef crypto_hash_sha256_state dtls_hash_ctx;
+typedef crypto_hash_sha256_state dtls_sha256_ctx;
+#define DTLS_HASH_CTX_SIZE sizeof(crypto_hash_sha256_state)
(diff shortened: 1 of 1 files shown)


obgm:
In the commit message: s/sha356/sha256/

mrdeep1:
Fixed - sorry, word blindness is still a challenge!

hmac.h
```diff
@@ -36,7 +36,19 @@ typedef sha256_context_t dtls_sha256_ctx;
...
 #define dtls_sha256_final(Buf,Ctx)        sha256_final((Ctx), (Buf))
 
-#else /* RIOT_VERSION */
+#elif defined ESPIDF_VERSION
+#include "sodium/crypto_hash_sha256.h"
+
+typedef crypto_hash_sha256_state dtls_hash_ctx;
+typedef crypto_hash_sha256_state dtls_sha256_ctx;
+#define DTLS_HASH_CTX_SIZE sizeof(crypto_hash_sha256_state)
+#define DTLS_SHA256_DIGEST_LENGTH (32)
```
obgm: I suggest using `crypto_hash_sha256_BYTES` here instead of the magic number.

mrdeep1: Fixed.

hmac.h
```diff
@@ -36,7 +36,19 @@ typedef sha256_context_t dtls_sha256_ctx;
 #define dtls_sha256_update(Ctx,Input,Len) sha256_update((Ctx), (Input), (Len))
 #define dtls_sha256_final(Buf,Ctx)        sha256_final((Ctx), (Buf))
 
-#else /* RIOT_VERSION */
+#elif defined ESPIDF_VERSION
```
obgm: The pre-processor symbol `ESPIDF_VERSION` is set in a different component (libcoap) and thus not very generic. 
I suggest testing for `ESP_PLATFORM` instead. Moreover, I presume the presence of libsodium should be checked as well.

mrdeep1: Fixed

PR #11, opened 2019-04-24
Separate out rijndael.c into primary and wrapper functions

mrdeep1:
Some platforms already provide the primary rijndael functions, but do not
have the specific TinyDTLS wrapper functions.

This change is to split out the primary (i.e. common) rijndael functions
(leave in rijndael.c) and the TinyDTLS wrapper functions (put into
rijndael_wrap.c) and removing the appropriate #ifdef DTLS_EXT_RIJNDAEL.

.gitignore:

Ignore .so library file

Makefile.in:
Makefile.tinydtls:
aes/Makefile.in:
aes/Makefile.riot:

Add in rijndael_wrap.[co] as appropriate.

aes/rijndael.c:

Remove the wrapper functions.

aes/rijndael_wrap.c: (New)

Add in the wrapper functions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ doc/Doxyfile
 doc/doxygen.out
 doc/html/
 libtinydtls.a
+libtinydtls.so
 tests/ccm-test
 tests/dtls-client
 tests/dtls-server
--- a/Makefile.in
+++ b/Makefile.in
@@ -39,7 +39,7 @@ install := cp
 
 # files and flags
 SOURCES:= dtls.c crypto.c ccm.c hmac.c netq.c peer.c dtls_time.c session.c dtls_debug.c dtls_prng.c
-SUB_OBJECTS:=aes/rijndael.o @OPT_OBJS@
+SUB_OBJECTS:=aes/rijndael.o aes/rijndael_wrap.o @OPT_OBJS@
 OBJECTS:= $(patsubst %.c, %.o, $(SOURCES)) $(SUB_OBJECTS)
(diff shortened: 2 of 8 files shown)


obgm:
```
gcc -Wall -std=c99 -pedantic -fPIC -DDTLSv12 -DWITH_SHA256 -DWITH_AES_DECRYPT  -c -o rijndael_wrap.o rijndael_wrap.c
rijndael_wrap.c: In function ‘rijndael_decrypt’:
rijndael_wrap.c:61:2: warning: implicit declaration of function ‘rijndaelDecrypt’; did you mean ‘rijndaelEncrypt’? [-Wimplicit-function-declaration]
  rijndaelDecrypt(ctx->dk, ctx->Nr, src, dst);
  ^~~~~~~~~~~~~~~
  rijndaelEncrypt
```

mrdeep1:
Actually, I think it still should be rijndaelDecrypt() as it was previously.  I have made rijndaelDecrypt() non static in rijndael.c and defined it in rijndael.h.

However, esp-idf does not expose (it is static)  rijndaelDecrypt(), but rijndaelDecrypt() is not used for TinyDTLS as far as I can tell.

obgm:
Correct. The issue here is that the prototype is missing from `rijndael.h`. We can either remove the decrypt part entirely or add the prototype to make `-DWITH_AES_DECRYPT` work properly.

mrdeep1:
I have made the prototype work (fixed rijndael.h) in the commit https://github.com/eclipse/tinydtls/commit/35cdb3d356d7969bb3e759a0ce95e692d1f7b683 pushed about 25 mins ago.

obgm:
Sorry, I have not seen that change before writing my comment. Looks good.

PR #15, opened 2019-07-10
use getrandom in place of srand on Linux

jvermillard:
rand() and srand() are not appropriate for use in cryptography
getrandom() is, c.f. man pages

Signed-off-by: Julien Vermillard <jvermillard@sierrawireless.com>

### DIFF
--- a/prng.h
+++ b/prng.h
@@ -32,6 +32,7 @@
 
 #ifndef WITH_CONTIKI
 #include <stdlib.h>
+#include <sys/random.h>
 
 /**
  * Fills \p buf with \p len random bytes. This is the default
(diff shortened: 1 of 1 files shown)


prng.h
```diff
@@ -40,14 +41,13 @@
...
-  while (len--)
-    *buf++ = rand() & 0xFF;
+  getrandom(buf, len, 0);
   return 1;
 }
 
 static inline void
 dtls_prng_init(unsigned short seed) {
-	srand(seed);
+  // nothing to do on linux
```
obgm: When you do this you should add `(void) seed` to avoid warnings caused by unused arguments.

obgm:
I would like to have this in the develop branch as well. There, a seed is read from `/dev/urandom` instead of using `getrandom()` in `dtls_prng()`.

obgm:
This has been superseded by 73d0a6c.

mrdeep1:
@jvermillard Can you please confirm that the fix now in the develop branch (https://github.com/eclipse/tinydtls/commit/73d0a6cea17aef2a027e0f8973c593c201f77f0c) does what you require.  If so, please close off this PR.

obgm:
Closing as resolved.

PR #18, opened 2019-09-02
Added Cmake support

JimmyBjorklund:
Added cmake support 
Signed-off-by: Jimmy Björklund <jimmy.bjorklund@hotmail.se>


CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+target_include_directories(tinydtls PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR})
+
+if( ${make_tests} )
+    add_subdirectory(tests)
+endif()
+
+
+if(${BUILD_SHARED_LIBS})
+    install(TARGETS    tinydtls       DESTINATION /usr/local/lib )    
+endif()
```
gocarlos, line 35: great that you provide this cmake script!

questions: why only installing when shared lib? missing also header files installation. :


```cmake

include(GNUInstallDirs)

install(
  TARGETS tinydtls
  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT lib
  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT lib)
install(
  DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
  COMPONENT dev
  FILES_MATCHING
  PATTERN "*.h")


```

gocarlos:
@obgm any chance we can make some progress here?

obgm:
@gocarlos Yes, I think it makes sense to go forward with this.

JKRhb:
What is missing for this PR to go forward? Does https://github.com/eclipse/tinydtls/pull/18#discussion_r404179481 need to be addressed first or could it already be merged?

obgm:
@JKRhb I think so.
The major obstacle is that we would need someone who is willing to work on this, i.e., keep it in sync and functionally equivalent to the configure/make mechanism.

JKRhb:
> @JKRhb I think so.

Okay, great :)

> The major obstacle is that we would need someone who is willing to work on this, i.e., keep it in sync and functionally equivalent to the configure/make mechanism.

I can try to do it, as I probably need to work with CMake in the near future anyway (maybe @JimmyBjorklund and @gocarlos can also have an eye on it?). For #126 to work with CMake, there also some adjustments needed which I would also try to add in.

AutoConf.cmake
```diff
@@ -0,0 +1,49 @@
...
+check_function_exists (memset         HAVE_MEMSET)
+check_function_exists (select         HAVE_SELECT)
+check_function_exists (socket         HAVE_SOCKET)
+check_function_exists (strdup         HAVE_STRDUP)
+check_function_exists (strerror       HAVE_STRERROR)
+check_function_exists (strnlen        HAVE_STRNLEN)
+check_function_exists (fls            HAVE_FLS)
+check_function_exists (vprintf        HAVE_VPRINTF)
+
+if( ${HAVE_STRING_H} AND ${HAVE_STRINGS_H} AND
```
boaks, line 38: The conditions seems to require "" around for the case, that a header is missing. e.g. "${HAVE_DLFCN_H}".

boaks:
> The major obstacle is that we would need someone who is willing to work on this, i.e., keep it in sync and functionally equivalent to the configure/make mechanism.

Zephyr requires to use cmake, otherwise it's hard to get it working. Therefore I started with a "dtls_config.h" for posix and the adapted single defines in order to compile it. Today I was spending some time in investigating, how the "autoconf" stuff is done in cmake/zephyr, and that part of this PR is very promising. (Beside of the TEST_BIG_ENDIAN this autoconf approach works also with zephyr.)

The amount of work is FMPOV not that much, it follows simple patterns. An additional define requires a `check_include_file` `check_function_exists` or `check_struct_has_member`. 

@JimmyBjorklund 
I guess, you're not longer interested, or?

If not, I would spend some time in the next 1-2 weeks to sync the stuff in "develop" with the stuff here.

JimmyBjorklund:
I have switched tasks so i am currently not looking to use this library. Just wanted to give something back for the good work every one else put in.

boaks:
Thanks for answer.
I will try to complete the CMake support. 
CMake support is also required for zephyr and your approach is already close to "work for zephyr" as well.

boaks:
I added a second commit with some changes.
One topic, I don't get, is the 

```
if(NOT PLATFORM)
   set(PLATFORM "posix" CACHE STRING "Choose platform." FORCE)
   set_property(CACHE PLATFORM PROPERTY STRINGS "contiki" "espidf" "posix" "riot" "zephyr")
endif()
```

I can't see, where this `PLATFORM` is used. Any hints, how that works?

obgm:
Closed as this PR has been superseded by #132. Thanks also to @gocarlos for contributing.

PR #21, opened 2019-12-05
dtls_time.h: fix CLOCK_SECOND redefined on FreeBSD

smlng:
when compiling RIOT on FreeBSD with TinyDTLS package it fails because the CLOCK_SECOND macro is already defined. This adds an `#undef` right before it redefinition to fix that.

### DIFF
--- a/dtls_time.h
+++ b/dtls_time.h
@@ -41,7 +41,11 @@
 
 #include <xtimer.h>
 
+/* this macro is already present on FreeBSD
+   which causes a redefine error otherwise */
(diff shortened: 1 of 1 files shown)


obgm:
If `CLOCK_SECOND` is already defined on a platform it should be used as-is instead of re-defining it. See, e.g., Contiki.

smlng:
changed to `#ifndef CLOCK_SECOND`

obgm:
Thank you!

Issue #20, opened 2019-10-02
client server example in RIOT failed with latest develop branch


pokgak:
I encountered this when testing PR #19 with the latest develop branch. Using the RIOT dtls-echo example, sending a packet from a client to a server will fail.

## How to reproduce

Apply the diff to the RIOT repository and run the dtls-echo example.

Use latest develop tinydtls branch on RIOT and enable debug by applying this diff:
```diff
diff --git a/examples/dtls-echo/Makefile b/examples/dtls-echo/Makefile
index 325078eac..91a51bc93 100644
--- a/examples/dtls-echo/Makefile
+++ b/examples/dtls-echo/Makefile
@@ -42,7 +42,7 @@ CFLAGS += -DDTLS_DEFAULT_PORT=$(DTLS_PORT)
 
 # NOTE: If not cipher suite is selected, DTLS_PSK is used by default.
 # This adds support for TLS_PSK_WITH_AES_128_CCM_8
-# CFLAGS += -DDTLS_PSK
+CFLAGS += -DDTLS_PSK
 # This adds support for TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8
 # CFLAGS += -DDTLS_ECC
 
@@ -52,7 +52,7 @@ TINYDTLS_LOG ?= 0
 
 # Enable this line for setting tinyDTLS in debug mode.
 # This overrides TINYDTLS_LOG to 6
-# CFLAGS += -DTINYDTLS_DEBUG
+CFLAGS += -DTINYDTLS_DEBUG
 
 # FIXME: This is a temporary patch
 # TinyDTLS <= 0.8.6 requires around 426 bytes in RAM.
diff --git a/pkg/tinydtls/Makefile b/pkg/tinydtls/Makefile
index e1a8483d0..802c5bc70 100644
--- a/pkg/tinydtls/Makefile
+++ b/pkg/tinydtls/Makefile
@@ -1,6 +1,6 @@
 PKG_NAME=tinydtls
 PKG_URL=https://github.com/eclipse/tinydtls.git
-PKG_VERSION=dcac93f1b38e74f0a57b5df47647943f3df005c2
+PKG_VERSION=494a40dfbb174930ca616e560532d52549736b42
 PKG_LICENSE=EPL-1.0,EDL-1.0
 
 CFLAGS += -Wno-implicit-fallthrough

```

<details>
<summary>Server log:</summary>
<pre><code>
> dtlss start
dtlss start
[...]

PR #22, opened 2019-12-12
dtls.c: fix error during handshake

smlng:
This fixes the issues mentioned in #20, I bisected and found it was introduces by https://github.com/eclipse/tinydtls/commit/68a1cdaff9e329e13ea59529f1eb61b05632c297

It seems that the length check is not correct, this is fixed here and the RIOT example work again.

Rational: the length check `(e + DTLS_HS_LENGTH + fragment_length) > msglen)` is wrong because below (see `dtls_hmac_update`) length is passed as `fragment_length - e` hence
```
e + DTLS_HS_LENGTH + (fragment_length - e) = DTLS_HS_LENGTH + fragment_length
``` 

Hope this makes sense. 

I also piggy-backed a minor code cleanup to make style consistent.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -384,7 +384,7 @@ dtls_create_cookie(dtls_context_t *ctx,
     return dtls_alert_fatal_create(DTLS_ALERT_HANDSHAKE_FAILURE);
 
   fragment_length = dtls_get_fragment_length(DTLS_HANDSHAKE_HEADER(msg));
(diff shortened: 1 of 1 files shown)


smlng:
@pokgak can you please verify that this fixes the issue, and maybe also check if my rational for the change makes sense 😄

obgm:
No, sorry, that is not a proper fix. Commit 7a0420b does the correct calculation but as said, there is still some other bug I am investigating.

smlng:
any how closing this one, as your fix does the trick. THX!

PR #31, opened 2020-05-30
add --disable-shared option

ffontaine:
Allow the user to disable shared library through --disable-shared.
Indeed, building of shared libraries is not supported by all toolchains
especially embedded ones

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


obgm:
Can you please rebase?

obgm:
Okay, thanks for the clarification. Here we go...

Issue #23, opened 2019-12-18
Renegotiation fails with test client/server application


pokgak:
Using the given `dtls-client` and `dtls-server` program in `tests/`, the server will fail when the client sends `client:renegotiate`.

Snippet from the *server* debug log ([full log](https://gist.github.com/pokgak/772520954c8873429a87507aedadb895)):

```
Dec 18 17:27:48 WARN decryption failed
Dec 18 17:27:48 INFO decrypt_verify() failed
```

After sending the command, the client also does not react anymore to new inputs from stdin.

obgm:
Thank you very much. With your detailed information, reproducing the issue was easy.

mrdeep1:
@pokgak This has been fixed in #41.  Please confirm that this fixes your issue.

pokgak:
Confirmed #41 fixes this issue. Handshake completed successfully and no more error when running `client:renegotiate`.

pokgak:
Closing because #41 is merged.

PR #32, opened 2020-06-01
Fix L used for nonce sizes, add extended master secret support (RFC7627)

mrdeep1:
Needed to enable EC interopability with GnuTLS and Californium.

crypto.c:

Fix L sizes broken by cbe1810f8c123b1b458dbb7f06acb5e02414edf3 reworking.

crypto.h:

Track extended_master_secret in dtls_handshake_parameters_t.

state.h:

Extended master secret hash info in dtls_hs_state_t.

global.h:

Add in #define for TLS_EXT_EXTENDED_MASTER_SECRET.

dtls.c:

Send out signature/hash algorithm in client hello so gnutls does not choose
SHA1. Verify signature/hash algorithm is supported in client hello.

Add in support for extended master secret as per RFC7627.

Stop sending out multiple alerts if bad extensions in client hello.

Allow server not to include client certificate type in server hello.

Define p before buf in stack so assert checking of buf overflow works.

Support servers that do not send back a hello verify request triggering
a second client hello.


obgm:
Thanks! As this PR deals with different things (bugfix vs. new feature), I would like to have the bugfix (L for nonce sizes) split off in a seperate PR.

dtls.c
```diff
@@ -112,7 +112,7 @@ memarray_t dtlscontext_storage;
 #define DTLS_HS_LENGTH sizeof(dtls_handshake_header_t)
 #define DTLS_CH_LENGTH sizeof(dtls_client_hello_t) /* no variable length fields! */
 #define DTLS_COOKIE_LENGTH_MAX 32
-#define DTLS_CH_LENGTH_MAX sizeof(dtls_client_hello_t) + DTLS_COOKIE_LENGTH_MAX + 12 + 26
+#define DTLS_CH_LENGTH_MAX sizeof(dtls_client_hello_t) + DTLS_COOKIE_LENGTH_MAX + 12 + 26 + 12
 #define DTLS_HV_LENGTH sizeof(dtls_hello_verify_t)
 #define DTLS_SH_LENGTH (2 + DTLS_RANDOM_LENGTH + 1 + 2 + 1)
 #define DTLS_CE_LENGTH (3 + 3 + 27 + DTLS_EC_KEY_SIZE + DTLS_EC_KEY_SIZE)
```
boaks, line 118: That `(3 + 3 + 27 + DTLS_EC_KEY_SIZE + DTLS_EC_KEY_SIZE)` evaluates to 97, but the message is 94.
I guess, that's a rudiment from that time, as a RPK certificate was implemented similar to x509 as list of certificates. But that was clarified away. So, please check, if one 3 should be removed.

dtls.c
```diff
@@ -914,7 +931,34 @@ static int verify_ext_ec_point_formats(uint8 *data, size_t data_length) {
...
+
+  /* length of sig_hash_algo list */
+  i = dtls_uint16_to_int(data);
+  data += sizeof(uint16);
+  if (i + sizeof(uint16) != data_length) {
+    dtls_warn("the list of the supported signature_algorithms should be tls extension length - 2\n");
+    return 1;
+  }
+
+  for (i = data_length - sizeof(uint8); i > 0; i -= sizeof(uint8)) {
```
obgm, line 948: This does not look right: We iterate data_length -1 times, and in each iteration, `data` is increased by two bytes, i.e., in the end, we have read 2 · (data_length - 1) bytes.

dtls.c
```diff
@@ -2016,8 +2081,8 @@ static int
 dtls_send_certificate_ecdsa(dtls_context_t *ctx, dtls_peer_t *peer,
 			    const dtls_ecdsa_key_t *key)
 {
-  uint8 buf[DTLS_CE_LENGTH];
   uint8 *p;
+  uint8 buf[DTLS_CE_LENGTH];
```
obgm, line 2085: I do not see why this needs to be re-ordered.

dtls.c
```diff
@@ -2106,8 +2171,8 @@ dtls_send_server_key_exchange_ecdh(dtls_context_t *ctx, dtls_peer_t *peer,
 {
   /* The ASN.1 Integer representation of an 32 byte unsigned int could be
    * 33 bytes long add space for that */
-  uint8 buf[DTLS_SKEXEC_LENGTH + 2];
   uint8 *p;
+  uint8 buf[DTLS_SKEXEC_LENGTH + 2];
```
obgm, line 2175: Why ist this re-ordered?

dtls.c
```diff
@@ -2198,8 +2263,8 @@ dtls_send_server_key_exchange_psk(dtls_context_t *ctx, dtls_peer_t *peer,
 static int
 dtls_send_server_certificate_request(dtls_context_t *ctx, dtls_peer_t *peer)
 {
-  uint8 buf[8];
   uint8 *p;
+  uint8 buf[8];
```
obgm, line 2267: Why is this re-ordered?

dtls.c
```diff
@@ -2443,8 +2516,8 @@ dtls_send_certificate_verify_ecdh(dtls_context_t *ctx, dtls_peer_t *peer,
 {
   /* The ASN.1 Integer representation of an 32 byte unsigned int could be
    * 33 bytes long add space for that */
-  uint8 buf[DTLS_CV_LENGTH + 2];
   uint8 *p;
+  uint8 buf[DTLS_CV_LENGTH + 2];
```
obgm, line 2520: Why is this re-ordered?

dtls.c
mrdeep1: Well spotted - yes, you are correct, DTLS_CE_LENGTH is too large by 3 bytes.  I think that this should be corrected in a separate PR.

dtls.c
```diff
@@ -813,14 +814,30 @@ calculate_key_block(dtls_context_t *ctx,
...
-	   DTLS_MASTER_SECRET_LENGTH);
+  if (handshake->extended_master_secret) {
+    unsigned char sha256hash[DTLS_HMAC_DIGEST_SIZE];
+
+    dtls_hash_finalize(sha256hash, &peer->handshake_params->hs_state.ext_hash);
+
+    dtls_prf(pre_master_secret, pre_master_len,
+  	     PRF_LABEL(extended_master), PRF_LABEL_SIZE(extended_master),
+	     sha256hash, sizeof(sha256hash)/2,
+	     &sha256hash[sizeof(sha256hash)/2], sizeof(sha256hash)/2,
```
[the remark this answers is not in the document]
mrdeep1: Corrected.

dtls.c
```diff
@@ -853,7 +870,7 @@ static int verify_ext_eliptic_curves(uint8 *data, size_t data_length) {
   data += sizeof(uint16);
   if (i + sizeof(uint16) != data_length) {
     dtls_warn("the list of the supported elliptic curves should be tls extension length - 2\n");
-    return dtls_alert_fatal_create(DTLS_ALERT_HANDSHAKE_FAILURE);
+    return 1;
```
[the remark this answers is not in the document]
mrdeep1: My mistake - I have reverted back these changes.  For some reason I thought that `dtls_alert_fatal_create()` was actually generating an alert rather than just returning a negative error.

dtls.c
mrdeep1: Corrected to below.  Result of a copy and past and partial update - sorry.
````
  for (i = data_length - sizeof(uint16); i > 0; i -= sizeof(uint16)) {
````

File: dtls.c (hunk 3 above)
Comment (mrdeep1) [reply dtls.c:2085]: Original code was of the form
````
  uint8 buf[some_size];
  uint8 *p;
  /* Start message construction at beginning of buffer. */
  p = buf;
  ... multiple instances of the next 2 lines
  memcpy(p, some_data, size)
  p += size;
  ...
  assert(p <= (buf + sizeof(buf)));
````
If the size of `buf `is too small, with the original ordering, after several `memcpy()`, `p` will point to (almost) the end of `buf `and the next `memcpy()` will actually overwrite `p` as it immediately follows `buf`.  I ran into this issue as `p` ended up being a large value and so the `assert()` did not trigger and other memory was corrupted.
Defining `p` before `buf` in the stack meant I caught my `buf` sizing error.

File: dtls.c (hunk 4 above)
Comment (mrdeep1) [reply dtls.c:2175]: See previous comment.

File: dtls.c (hunk 5 above)
Comment (mrdeep1) [reply dtls.c:2267]: See previous comment

File: dtls.c (hunk 6 above)
Comment (mrdeep1) [reply dtls.c:2520]: See previous comment.

**Comment by mrdeep1** [comment]
Code changes pushed.

File: dtls.c (hunk 1 above)
Comment (mrdeep1) [reply dtls.c:118]: PR #40 created to fix this.

File: dtls.c (hunk 3 above)
Comment (boaks) [reply dtls.c:2085]: Hell!
if you reorder that, I hope you don't kill your call-stack! Otherwise a return may end in "no-mans-land".

Comment (obgm) [reply dtls.c:2085]: Well, the idea is to fix potential buffer overflows eventually.

Comment (mrdeep1) [reply dtls.c:2085]: Hmm, which ever way round these are defined, an overflow of `buf` will cause issues.

I have just disassembled the x86 code with them being defined in either order - in both cases `p` is higher in the stack than `buf ` for the couple of functions I checked - so compiler is doing some re-arranging here (likely to be down to 8 byte alignment and making use of non-aligned space).  I will revert out these `p `/ `buf `re-ordering changes as they appear to be making no difference and we are pretty confident that the `buf `sizes are OK.

**Comment by mrdeep1** [comment]
Code pushed with reverted out `p `/ `buf `re-ordering

File: dtls.c (hunk 3 above)
Comment (obgm) [reply dtls.c]: Yes, it is on the optimizer to decide. In theory, `p` could even end up in a register. Anyway, if you compiled with `-O0`, you should see the order you have specified in the code.

Comment (boaks) [reply dtls.c]: Decades ago I used macro's and a small set of functions to but a "mark" around buffers and verify that. That fails to detect "jumps in buffer" (e.g. p *= 128; *p= 0). But helps to locate a couple of violations. It's too long ago, that I was really "up-to-date" in C, so maybe, in the meantime, compilers offers such things.

But tricks like that "order of buffer and point", didn't work in my experience of the past.

Comment (mrdeep1) [reply dtls.c]: It is likely I was working with no optimization as I was debugging the code when I stumbled into this issue.

File: dtls.c [hunk 9]
```diff
@@ -2643,16 +2713,45 @@ dtls_send_client_hello(dtls_context_t *ctx, dtls_peer_t *peer,
...
   assert((buf <= p) && ((unsigned int)(p - buf) <= sizeof(buf)));
 
-  if (cookie_length != 0)
-    clear_hs_hash(peer);
+  clear_hs_hash(peer);
 
   return dtls_send_handshake_msg_hash(ctx, peer, &peer->session,
 				      DTLS_HT_CLIENT_HELLO,
-				      buf, p - buf, cookie_length != 0);
+				      buf, p - buf, 1);
```
obgm, line 2754: I am having a hard time wrapping my head around this change. It looks as if you decided that independent of the cookie's presence a new hash calculation is started at this point

mrdeep1: Not all servers return a hello_verify (gnutls did not until I fixed that in libcoap).  As there was no second client_hello, client_hello never never made it into the hash.  This is the reason why I now always reset the hash in client_hello so that we only get one client_hello in the hash.

obgm: So far, that sounds reasonable. I was just wondering if this might affect other invocations of `dtls_send_client_hello()` where cookie_length is explicitly set to 0, e.g., in `dtls_renegotiate()`?

mrdeep1: Interesting question.  During this code development, I did a quick test of #23 to check I was not making that worse - and found that it worked.  However, when I rechecked at the end of the coding/testing, #23 was not working again, but I could not get to to work again - a change like this may have made a difference here - I will try to test that further to understand what is really happening there in #23

mrdeep1: This does not break dtls_renegotiation().  #23 is now fixed in PR #41

PR #33, opened 2020-06-08
dtls.h: Remove readbuf from dtls_context_t as it is unused

mrdeep1:
This allows RAM to be freed up.

Signed-off-by: Jon Shallow supjps-libcoap@jpshallow.com


obgm:
Weird to have something like this being undetected for such a long time.

PR #34, opened 2020-06-10
crypto.c: Fix L used for nonce sizes

mrdeep1:
Fix L sizes broken by cbe1810f8c123b1b458dbb7f06acb5e02414edf3 reworking.

### DIFF
--- a/crypto.c
+++ b/crypto.c
@@ -581,8 +581,8 @@ dtls_encrypt(const unsigned char *src, size_t length,
 	     const unsigned char *aad, size_t la)
 {
   /* For backwards-compatibility, dtls_encrypt_params is called with
-   * M=8 and L=2. */
-  const dtls_ccm_params_t params = { nonce, 8, 2 };
+   * M=8 and L=3. */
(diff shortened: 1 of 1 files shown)


chrysn:
This appears to be missing a change in the documentation. The `dtls_encrypt` doc comment still says that "dtls_encrypt() always sets M=8, L=2" and similar.

L=3 is an odd choice (message lengths of 16bit are *encoded* as value 3, but everywhere else it appears to be consistently used as value and not encoding). If the legacy usage of dtls_encrypt / dtls_decrypt is really meant to be with 3-byte long message lengths, it may help to have a comment around the param-less functions along the lines of

>     * \deprecated dtls_encrypt() always sets M=8, L=3, which is a value only in use by
>     *             old tinydtls versions. New code should use dtls_encrypt_params() instead,
>     *             and set M=8, L=3 in its params for compatibility with those versions, or any
>     *             of the more widespread values M=8, L=2, or M=16, L=8, or combinations
>     *             thereof.

(If that's really how that 3 came to be; otherwise, docs could still need an update but the text would be different.)

PR #40, opened 2020-06-27
dtls.c: Fix size of buffer for dtls_send_certificate_ecdsa()

mrdeep1:
As per #32 DTLS_CE_LENGTH evaluates to 97, but only 94 bytes are needed.

Remove DTLS_CE_LENGTH and define output buffer size for 
dtls_send_certificate_ecdsa() for clarity to be
  uint8 buf[sizeof(uint24) + DTLS_EC_SUBJECTPUBLICKEY_SIZE];

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -115,7 +115,6 @@ memarray_t dtlscontext_storage;
 #define DTLS_CH_LENGTH_MAX sizeof(dtls_client_hello_t) + DTLS_COOKIE_LENGTH_MAX + 12 + 26
 #define DTLS_HV_LENGTH sizeof(dtls_hello_verify_t)
 #define DTLS_SH_LENGTH (2 + DTLS_RANDOM_LENGTH + 1 + 2 + 1)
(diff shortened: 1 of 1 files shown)


obgm:
The assessment in PR #32 is correct that this is a relic from a time where an RPK was treated as certificate list which would have required an additional 24-bit field. When changing this in 2015, I have missed the additional three bytes in the calculation.

I think that we should employ a different fix to make the code more readable:
* `DTLS_CE_LENGTH` is used only to set the buffer size in `dtls_send_certificate_ecdsa(), and
* `dtls_send_certificate_ecdsa()` already writes `DTLS_EC_SUBJECTPUBLICKEY_SIZE` as the certificate length which is
  actually `DTLS_CE_LENGTH` - 3 (= sizeof uint24).

Therefore, I suggest deleting the definition of `DTLS_CE_LENGTH` entirely and change the declaration of `buf` in `dtls_send_certificate_ecdsa()` to

```c
  uint8 buf[sizeof(uint24) + DTLS_EC_SUBJECTPUBLICKEY_SIZE];
```

mrdeep1:
Code changes pushed.

PR #41, opened 2020-06-30
dtls.c: Fix renegotiate

### LINKED ISSUE #23: Renegotiation fails with test client/server application

mrdeep1:
The renegotiate Client Hello had the record layer set to DTLS 1.0 when
it should have been DTLS 1.2.

Only when epoch == 0 set it to DTLS 1.0.

This fix does not include support for RFC5746 Renegotiation Indication
Extension

This fixes #23 

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -1595,11 +1595,15 @@ dtls_send_multi(dtls_context_t *ctx, dtls_peer_t *peer,
   /* Signal DTLS version 1.0 in the record layer of ClientHello and
    * HelloVerifyRequest handshake messages according to Section 4.2.1
    * of RFC 6347.
+   *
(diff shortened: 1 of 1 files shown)


obgm:
Great, thank you very much!
And thanks to @pokgak for checking.

PR #44, opened 2020-07-23
configure.ac: Fix CUNIT spurious bracket

mrdeep1:
configure.ac:
Remove spurious ] from the CUNIT test

README:

Include autoreconf to handle configure.ac updates.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -92,7 +92,7 @@ if test "x$build_tests" = "xyes"; then
     PKG_CHECK_MODULES([CUNIT],
                       [cunit],
                       [have_cunit=yes
-                       AC_DEFINE(HAVE_LIBCUNIT, [1], [Define if the system has libcunit])]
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -93,7 +93,7 @@ if test "x$build_tests" = "xyes"; then
                       [cunit],
                       [have_cunit=yes
                        AC_DEFINE(HAVE_LIBCUNIT, [1], [Define if the system has libcunit])]
-                       AC_DEFINE(TEST_INCLUDE, [1], [Define to include test wrappers for static functions])],
+                       AC_DEFINE(TEST_INCLUDE, [1], [Define to include test wrappers for static functions]),
```
obgm, line 96: I think that the `]` is correct in this line. The spurious `]` is in the previous line.

mrdeep1: Corrected and pushed.

PR #49, opened 2021-01-25
dtls_contiki-ng: add support to contiki-ng

lutgaru:
Minor changes to add support as a submodule of contiki-ng.

Currently contiki-ng is using an outdated standalone version of tinydtls as a submodule, which causes problems that have already been resolved in the develop branch, such as (https://github.com/obgm/libcoap/issues/160).

These changes will allow the development branch to be used as a submodule of contiki-ng.

Signed-off-by: lutgaru lutgaru@gmail.com

### DIFF
--- a/session.h
+++ b/session.h
@@ -23,13 +23,18 @@
 #include "global.h"
 
 #ifdef WITH_CONTIKI
-#include "ip/uip.h"
+#ifdef BUILD_WITH_COAP
+#include "coap-endpoint.h"
+typedef coap_endpoint_t session_t;
+#else /*BUILD_WITH_COAP*/
+#include "uip.h"
 typedef struct {
   unsigned char size;
-  uip_ipaddr_t addr;
+  uip_ipaddr_t ipaddr;
   unsigned short port;
   int ifindex;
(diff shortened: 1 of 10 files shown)


lutgaru:
sorry I was missing something important, apparently a new makefile is not necessary, but if it was necessary to fix the warnings of the test files since contiki cannot avoid building them.

session.h
```diff
@@ -23,13 +23,8 @@
...
 
 #ifdef WITH_CONTIKI
-#include "ip/uip.h"
-typedef struct {
-  unsigned char size;
-  uip_ipaddr_t addr;
-  unsigned short port;
-  int ifindex;
-} session_t;
+#include "coap-endpoint.h"
```
obgm, line 26: As tinydtls is not CoAP-specific in any way, I would rather not create this dependency.

lutgaru: it's true, I just corrected this adding it only in the case that it is being built in CoAP where the pointer does have to be the same.

lutgaru:
I need to review this again, I'll upload it when everything is ok

PR #53, opened 2021-03-11
close alert: Make sending close notify alert as a warning, not fatal …

### LINKED ISSUE #51: CLOSE_NOTIFY - level FATAL instead of WARNING?
Using tinydtls  "develop" branch as binding for libcoap, I receive the `CLOSE_NOTIFY` alert with level `FATAL`. 
AFAIK, that should be `WARNING`.

So, is the `FATAL` intended?

mrdeep1:
…type

While not defined which way to send a close notify alert, most implementations
have gone with sending it as type warning, not fatal.  This brings TinyDTLS
into line.

Fixes #51

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -1732,7 +1732,8 @@ dtls_close(dtls_context_t *ctx, const session_t *remote) {
   peer = dtls_get_peer(ctx, remote);
 
   if (peer) {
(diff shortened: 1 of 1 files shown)


boaks:
Thanks, with this PR my tests are no receiving the close_notify as warning.

PR #60, opened 2021-03-24
dtls.c: Correctly handle missing packets checking

### LINKED ISSUE #57: Undefined Behavior Due to Over-shift of Record Sequence Number
Hello again,

The bug occurs in the part that is responsible for checking the validity of the `record sequence number`. This bug is found by employing symbolic execution technique. When a packet is being processed (in `dtls_handle_message()`), if a peer is initialized (`dtls_peer_t *peer`), TinyDTLS checks for the validity of the `record sequence number`.  In the case that (`pkt_seq_nr > security->cseq.cseq`), we have shift operations as the [following](https://github.com/eclipse/tinydtls/blob/master/dtls.c#L3774):
```
security->cseq.bitfield <<= (pkt_seq_nr-security->cseq.cseq);  
security->cseq.bitfield |= 1<<((pkt_seq_nr-security->cseq.cseq)-1); 
```
we know that `cseq.bitfield` is an unsigned long (64-bit) and 1 is an unsigned integer (32-bit). Therefore, if the value of `pkt_seq_nr-security->cseq.cseq` is greater equal to 64 there is an over-shift for the first statement and for the values of `pkt_seq_nr-security->cseq.cseq` greater equal to 32 there is an over-shift for the second statement. Finally, since `pkt_seq_nr-security->cseq.cseq` refers to the `Record Sequence Number` in a dtls record, for the `Record Sequence Number > 32`, we will have an over-shift. 
I have attached the means to reproduce the mentioned bug. To do so, after downloading the suite, in `tinydtls-master-witness/tests`, execute the script `setupserver.sh` to compile TinyDTLS with `Undefined Behavior Sanitizor`. Note that, I have de-randomized TinyDTLS for this demonstration and the patch is available in` tinydtls-master-witness/tests/patches`. Successful execution of `setupserver.sh` will run the `dtls-server` on port 20220. Now, executing `./reproduce.sh` in `tinydtls-master-witness/tests` should result in the following run-time error from the sanitizer:

image

I hope I could explain it understandably. I appreciate the confirmation of the bug.
[tinydtls-master-witness.zip](https://github.com/eclipse/tinydtls/files/6189172/tinydtls-master-witness.zip)

Best,
Hooman Asadian

mrdeep1:
Make sure that bitfield is correctly used for tracking missing packets and
that a large block of missing packets do not cause an over shift.

Fixes #57.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -4039,14 +4039,15 @@ dtls_handle_message(dtls_context_t *ctx,
           data_length = decrypt_verify(peer, msg, rlen, &data);
           if (data_length) {
             security->cseq.cseq = 0;
-            security->cseq.bitfield = -1;
+            /* bitfield. B0 last seq seen.  B1 seq-1 seen, B2 seq-2 seen etc. */
+            security->cseq.bitfield = 1;
           }
         } else if (pkt_seq_nr == security->cseq.cseq) {
           dtls_info("Duplicate packet arrived (cseq=%" PRIu64 ")\n", security->cseq.cseq);
           return 0;
         } else if ((int64_t)(security->cseq.cseq-pkt_seq_nr) > 0) { /* pkt_seq_nr < security->cseq.cseq */
           if (((security->cseq.cseq-1)-pkt_seq_nr) < 64) {
-              if(security->cseq.bitfield & (1<<((security->cseq.cseq-1)-pkt_seq_nr))) {
+              if(security->cseq.bitfield & ((uint64_t)1<<((security->cseq.cseq-1)-pkt_seq_nr))) {
                 dtls_info("Duplicate packet arrived (bitfield)\n");
                 /* seen it */
                   return 0;
@@ -4054,7 +4055,7 @@ dtls_handle_message(dtls_context_t *ctx,
                 dtls_info("Packet arrived out of order\n");
                 data_length = decrypt_verify(peer, msg, rlen, &data);
                 if(data_length > 0) {
-                  security->cseq.bitfield |= (1<<((security->cseq.cseq-1)-pkt_seq_nr));
+                  security->cseq.bitfield |= ((uint64_t)1<<((security->cseq.cseq-1)-pkt_seq_nr));
                 }
               }
           } else {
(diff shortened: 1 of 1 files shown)


boaks:

dtls.c
```diff
@@ -3958,33 +3958,42 @@ dtls_handle_message(dtls_context_t *ctx,
...
                 if(data_length > 0) {
-                  security->cseq.bitfield |= (1<<((security->cseq.cseq-1)-pkt_seq_nr));
+                  security->cseq.bitfield |= ((uint64_t)1<<((security->cseq.cseq-1)-pkt_seq_nr));
                 }
               }
           } else {
             dtls_info("Packet from before the bitfield arrived\n");
               return 0;
           }
+        } else if (pkt_seq_nr-security->cseq.cseq > 63) {
```
obgm, line 3984: As the only difference between this new conditional part and the following else is the `cseq.bitfield` value written, I suggest removing these new lines 3984—3990 and instead change the `else` part as follows to avoid code duplication:

```c
           data_length = decrypt_verify(peer, msg, rlen, &data);
           if(data_length > 0) {
             security->cseq.bitfield <<= (pkt_seq_nr-security->cseq.cseq);
-            security->cseq.bitfield |= 1<<((pkt_seq_nr-security->cseq.cseq)-1);
+            if (pkt_seq_nr - security->cseq.cseq > 63) {
+              /* reset bitfield if new packet number is beyond its boundaries */
+              security->cseq.bitfield = 0;
+            }
+            security->cseq.bitfield |= 1;
             security->cseq.cseq = pkt_seq_nr;
             dtls_debug("new packet arrived with seq_nr: %" PRIu64 "\n", pkt_seq_nr);
             dtls_debug("new bitfield is               : %" PRIx64 "\n", security->cseq.bitfield);
```

boaks: If `security->cseq.bitfield = 0;` would `security->cseq.bitfield <<= (pkt_seq_nr-security->cseq.cseq);` in a `else` fit better?

```
if (pkt_seq_nr - security->cseq.cseq > 63) {
   /* reset bitfield if new packet number is beyond its boundaries */
   security->cseq.bitfield = 0;
} else {
   /* shift bitfield */
   security->cseq.bitfield <<= (pkt_seq_nr-security->cseq.cseq);
}
```

obgm: Good point, agreed.

obgm: (Apologies for the missing LF in the code.)

boaks: > (Apologies for the missing LF in the code.)

Comments can be edited in github using the "..." left to the smile :-).

obgm: Yes, I know the basics of the GH interface. But I would have had to generate a new diff which would have been a window change. Too much GUI interaction...

mrdeep1: Taken this version and updated accordingly.  Code pushed.

boaks:
My feeling on analyzing that code a couple of days ago was, that `((security->cseq.cseq-1)-pkt_seq_nr)` or  `(security->cseq.cseq - pkt_seq_nr)` is used all over. 

Maybe, if  `(pkt_seq_nr - security->cseq.cseq)` is calculated once and then the result is used for the comparisons, that "shrinks" not only the code, but also the irritations reading the same or similar expression so many times.

obgm:
That is a valid point.  As I am fine with the latest version in commit 4407147 we can do the merge first and cleanup in a separate PR or do it as a second commit in this PR. @mrdeep1?

mrdeep1:
I can look at this later tonight, and come up with something.  Otherwise, I am neutral as to whether to go with https://github.com/eclipse/tinydtls/commit/4407147f742f5f91851c6f6cf9f2c84b7f746ba9

mrdeep1:
I did not get a proper look at this - but there is both `security->cseq.cseq - pkt_seq_nr` and` pkt_seq_nr - security->cseq.cseq` so this could be done with a common ssize_t.  Perhaps best to merge https://github.com/eclipse/tinydtls/commit/4407147f742f5f91851c6f6cf9f2c84b7f746ba9 and visit this later.

boaks:
> Perhaps best to merge 4407147 and visit this later.

Agreed.

> I did not get a proper look at this - but there is both security->cseq.cseq - pkt_seq_nr and pkt_seq_nr - security->cseq.cseq so this could be done with a common ssize_t. 

AFAIK, the relevant range is from -64 to 64, larger or smaller values are processed as the -64 or 64.

obgm:
Okay, thanks @boaks and @mrdeep1  for sorting this out.

PR #61, opened 2021-03-24
dtls.c: Check length of data for ServerHello

### LINKED ISSUE #55: Buffer Over-read Affecting check_server_hello()
Hello again,

The following over-read happens when you send a fragmented Server Hello message to the client. The bug is found employing symbolic execution technique. Suppose you have a `Server Hello` message with the following values for certain fields:

- `length` in the record layer = 46
- `length` in the handshake layer = 0
- `fragment length` = 128

When we are handling the message (in `dtls_handle_message()`), `data_length = decrypt_verify(peer, msg, rlen, &data)` is called. It returns a pointer to the first offset of the handshake layer (`uint8 *data`) and the real size of the handshake layer (`data_length` = 46). Later inside `check_server_hello(ctx, peer, data, data_length)`, Tiny tries to read different fields of the `Server Hello` message. It happens by advancing the pointer `data` and deducting the `data_length`. As I previously mentioned, the `Server Hello` is fragmented and it contains only up to the `random` field (`handshake header = 12` bytes + `handshake version = 2` bytes + `random = 32` bytes). Therefore, after advancing the pointer (`data`) for 46 bytes, Tiny tries to skip the `Session ID` by calling the [Macro](https://github.com/eclipse/tinydtls/blob/feature/handshake_fragmentation/dtls.c#L2765) (`SKIP_VAR_FIELD(data, data_length, uint8)`). However, Tiny already has advanced the pointer (`data`) for 46 bytes, and the pointer (`data`) is currently pointing to invalid memory. Therefore, here we have a one-byte over-read. I have added the assertion (`assert(data_length > 0)`) before `SKIP_VAR_FIELD(data, data_length, uint8)` to reveal the invalid read. For reproducing the bug, I have written a small UDP server that plays the role of a DTLS-server. 
I have attached the means to reproduce the mentioned bug. To do so, after downloading the suite, in tinydtls-witness/tests, execute the script `setupserver.sh` to compile TinyDTLS and `simple-udp-server`. Note that, I have de-randomized TinyDTLS for this demonstration and the patch is available in `tinydtls-witness/tests/patches`. Successful execution of `setupserver.sh` will run `simple-udp-server` on port 20220. Now, if you execute `reproduce.sh`, it will initiate a handshake with `simple-udp-server` and when it gets the Server Hello, the assertion that I added will fail.

image

I hope I could explain it understandably. I appreciate the confirmation of the reported bugs.

Best,
Hooman Asadian
[tinydtls-witness.zip](https://github.com/eclipse/tinydtls/files/6182299/tinydtls-witness.zip)

mrdeep1:
If insufficient data, alert out.

Fixes #55.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * Copyright (c) 2011-2020 Olaf Bergmann (TZI) and others.
+ * Copyright (c) 2011-2021 Olaf Bergmann (TZI) and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompanies this distribution.
@@ -2763,17 +2763,19 @@ check_server_hello(dtls_context_t *ctx,
 {
   dtls_handshake_parameters_t *handshake = peer->handshake_params;
 
-  /* This function is called when we expect a ServerHello (i.e. we
-   * have sent a ClientHello).  We might instead receive a HelloVerify
-   * request containing a cookie. If so, we must repeat the
-   * ClientHello with the given Cookie.
+  /*
+   * Check we have enough data for the ServerHello
+   *   2 bytes for the version number
+   *   1 byte for the session id length
+   *   2 bytes for the selected cipher suite
+   *   1 byte null compression
    */
-  if (data_length < DTLS_HS_LENGTH + DTLS_HS_LENGTH)
+  if (data_length < DTLS_HS_LENGTH + 2 + DTLS_RANDOM_LENGTH + 1 + 2 + 1) {
+    dtls_alert("Insufficient length for ServerHello\n");
     return dtls_alert_fatal_create(DTLS_ALERT_DECODE_ERROR);
+  }
 
   update_hs_hash(peer, data, data_length);
-
-  /* FIXME: check data_length before accessing fields */
 
   /* Get the server's random data and store selected cipher suite
    * and compression method (like dtls_update_parameters().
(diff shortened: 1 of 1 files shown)


bathooman (approved):
The pull request successfully resolves the issue.

boaks:

dtls.c
```diff
@@ -2766,18 +2766,14 @@ check_server_hello(dtls_context_t *ctx,
...
   dtls_handshake_parameters_t *handshake = peer->handshake_params;
 
-  /* This function is called when we expect a ServerHello (i.e. we
-   * have sent a ClientHello).  We might instead receive a HelloVerify
-   * request containing a cookie. If so, we must repeat the
-   * ClientHello with the given Cookie.
-   */
-  if (data_length < DTLS_HS_LENGTH + DTLS_HS_LENGTH)
+  /* Check we have enough data for the ServerHello */
+  if (data_length < DTLS_HS_LENGTH + 2 + DTLS_RANDOM_LENGTH + 1 + 2 + 1) {
```
obgm, line 2770: This test solves only part of the issue as the session id might not be empty. To handle that case, the data access after the `SKIP_VAR_FIELD()` for the session id must be protected as well.

dtls.c
```diff
@@ -2766,18 +2766,14 @@ check_server_hello(dtls_context_t *ctx,
 {
   dtls_handshake_parameters_t *handshake = peer->handshake_params;
 
-  /* This function is called when we expect a ServerHello (i.e. we
-   * have sent a ClientHello).  We might instead receive a HelloVerify
-   * request containing a cookie. If so, we must repeat the
-   * ClientHello with the given Cookie.
-   */
-  if (data_length < DTLS_HS_LENGTH + DTLS_HS_LENGTH)
+  /* Check we have enough data for the ServerHello */
```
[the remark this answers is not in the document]
mrdeep1: Comment updated

dtls.c
mrdeep1: Good catch.  Length rechecked after `SKIP_VAR_FIELD()` and code pushed.

PR #62, opened 2021-03-24
dtls.c: Fix handling ClientHello if no peer and fragment

### LINKED ISSUE #54: Buffer Over-read Affecting Fragmentation Feature
Hello,

I have tried to submit this through Eclipse Bugzilla. However, apparently, the [dedicated page](https://bugs.eclipse.org/bugs/enter_bug.cgi?product=tinydtls) for TinyDTLS has been removed or I did not have access to it. The following over-read has been found by using symbolic execution technique. Suppose you have a `Client Key Exchange` message with the following values for the mentioned fields:
 
- `length` in the record layer = 25 
- `length` in the handshake layer = 51 
- `fragment length` = 16
- `fragment offset `= 0

When we are handling the message (in `dtls_handle_message()`), `data_length = decrypt_verify(peer, msg, rlen, &data)` is called. It returns a pointer to the first offset of the handshake layer (`uint8 *data`) and the real size of the handshake layer (`data_length = 25`). Later, as we have `packet_length=51 > fragment_length=16`, we start the reassembling process. the message also pass `fragment_offset=0 + fragment_length=16 <= packet_length=51` check. Finally, after allocating memory on the heap (`peer->handshake_params->reassemble_buf->data`), when we are going to copy the message ([memcpy](https://github.com/eclipse/tinydtls/blob/feature/handshake_fragmentation/dtls.c#L3716)) into the newly allocated buffer, we use the combination of `fragment offset` and `fragment length` to derive the size of the handshake message and the pointer (`data`) as it points to the first offset of the handshake layer.  Let me remind you that the valid range for `data` is 25 (`=data_length`). However, in this scenario, the size (fragment_length=16 + handshake_header_size=12) is 28. Therefore, there is a 3-byte overread happening. 
I have attached the means to reproduce the mentioned bug. To do so, after downloading the suite, in `tinydtls-witness/tests`, execute the script `setupserver.sh` to compile `TinyDTLS`. Note that, I have de-randomized TinyDTLS for this demonstration and the patch is available in `tinydtls-witness/tests/patches`. Successful execution of  `setupserver.sh` will run `dtls-server` on port 20220. Now, executing `./reproduce.sh` in `tinydtls-witness/tests` should cause `dtls-server` to crash. 
Screenshot from 2021-03-19 18-44-03

The malformed `Client Key Exchange` can be found in  `tinydtls-witness/tests/testcases`.

I hope I could explain it understandably. 
[tinydtls-witness.zip](https://github.com/eclipse/tinydtls/files/6173135/tinydtls-witness.zip)

Best wishes,
Hooman Asadian

mrdeep1:
If a fragmented ClientHello is received with no peer, do not handle it.

Note that this is a fix to feature/handshake_fragmentation branch to fix #54


dtls.c
```diff
@@ -3676,6 +3676,11 @@ handle_handshake(dtls_context_t *ctx, dtls_peer_t *peer, session_t *session,
   size_t fragment_offset = dtls_uint24_to_int(hs_header->fragment_offset);
 
   if (packet_length > fragment_length){
+    if (!peer || !peer->handshake_params) {
```
boaks, line 3679: I'm not sure, which coding style is used, but `!peer` seems to operate on a pointer.

I would prefer 

```c
if (data != NULL) {
```

as also used in that file (around line 1547).

mrdeep1: `!peer `is used 24 times throughout` dtls.c` as opposed to `peer == NULL`.  Likewise, there a a lot of `if (peer)`.  Line 3679 was copied from line 3756.

Yes, the coding style can be updated throughout, but this will mean a lot of changes.

boaks: Ah, OK.
I should have spent more time in to investigation before commenting.

Issue #59, opened 2021-03-23
Buffer Over-read Affecting dtls_create_cookie()


bathooman:
Hello again,

This bug occurs when the server receives a malformed `Client Hello` to initialize a handshake and it affects the master branch. This bug is found by employing symbolic execution technique. Suppose you have a malformed `Client Hello` where the `length` field in the handshake layer is 0. When the mentioned `Client Hello` is received by the server, `dtls_verify_peer()` calls `dtls_create_cookie()` for cookie generation. Inside `dtls_create_cookie()`, TinyDTLS defines `e` which is unsigned (`size_t`). Later, when  we call the following [update function](https://github.com/eclipse/tinydtls/blob/master/dtls.c#L356):

```
dtls_hmac_update(&hmac_context, 
		   msg + DTLS_HS_LENGTH + e,
		   dtls_get_fragment_length(DTLS_HANDSHAKE_HEADER(msg)) - e);
```
the macro `dtls_get_fragment_length(DTLS_HANDSHAKE_HEADER(msg))` will return 0. Therefore, since `e` is unsigned, `dtls_get_fragment_length(DTLS_HANDSHAKE_HEADER(msg)) - e` will wrap around and evaluates to a very huge number(e.g 18446744073709551580). This huge value, in turns will cause a memory over-read in the [memcpy](https://github.com/eclipse/tinydtls/blob/master/sha2/sha2.c#L571) in sha2.c.
I have attached the means to reproduce the mentioned bug. To do so, after downloading the suite, in `tinydtls-master-witness/tests`, execute the script `setupserver.sh` to compile TinyDTLS. I have not changed anything in the code for the sake of this demonstration. Successful execution of setupserver.sh will run the dtls-server on port 20220. Now, executing `./reproduce.sh `in `tinydtls-master-witness/tests` should crash the server as shown in the following figure:

![image](https://user-images.githubusercontent.com/3464382/112192384-32362100-8c07-11eb-8a50-3617f6553833.png)

I hope I could explain it understandably and I appreciate your confirmation of the bug. 
[tinydtls-master-witness.zip](https://github.com/eclipse/tinydtls/files/6191750/tinydtls-master-witness.zip)

Best,
Hooman Asadian

mrdeep1:
[...]

PR #64, opened 2021-03-25
dtls.c: Reject any fragmented handshake until fragmentation supported

mrdeep1:
Until feature/handshake_fragmentation is merged, reject any fragmented
handshake otherwise fragmented handshakes may cause a DoS.

Addresses issue raised in #59.


boaks:

PR #66, opened 2021-03-27
dtls-server.c: Trap SIGINT to help with valgrind

mrdeep1:

### DIFF
--- a/tests/dtls-server.c
+++ b/tests/dtls-server.c
@@ -20,6 +20,8 @@
 
 #define DEFAULT_PORT 20220
 
+static dtls_context_t *the_context = NULL;
+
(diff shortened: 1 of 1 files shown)


tests/dtls-server.c
```diff
@@ -324,6 +332,11 @@ main(int argc, char **argv) {
 
   dtls_init();
 
+  if (signal(SIGINT, dtls_handle_signal) == SIG_ERR) {
```
obgm, line 335: As `signal()` is explicitly deprecated, I suggest using `sigaction()` instead.

mrdeep1: Updated and code pushed

PR #67, opened 2021-03-27
github: Add in build and test workflow

mrdeep1:
Unfortunately out of path builds do not work at present.


mrdeep1:
Hmm - seems that github workflows are not enabled.

PR #80, opened 2021-05-26
[ECC] ecc.c: fix build with -Werror=sign-compare

benpicco:
`count` is unsigned while `result` and `idx` are both signed.
This causes a build failure with `-Werror`.

Since both variables can never assume negative values, just make them `unsisgned` as well.

### DIFF
--- a/ecc/ecc.c
+++ b/ecc/ecc.c
@@ -310,8 +310,8 @@ static void fieldModO(const uint32_t *A, uint32_t *result, uint8_t length) {
  * @return 1 if all elements in @val are zero, 0 otherwise.
  */
 static int is_zero(const uint32_t* val, size_t count) {
-  int result = 0;
-  int idx;
+  unsigned result = 0;
+  unsigned idx;
 
   for (idx = 0; idx < count; idx++) {
     result += val[idx] == 0;


mrdeep1:
Strictly speaking, `int `should have been replaced by `size_t `to match the calling parameter.  There are other code tidy up needed, in particular in the tests - see #81

ecc/ecc.c
```diff
@@ -310,8 +310,8 @@ static void fieldModO(const uint32_t *A, uint32_t *result, uint8_t length) {
  * @return 1 if all elements in @val are zero, 0 otherwise.
  */
 static int is_zero(const uint32_t* val, size_t count) {
-  int result = 0;
-  int idx;
+  unsigned result = 0;
+  unsigned idx;
```
obgm, line 314: I would like to see `size_t` here for `result` and `idx` even if this is not really necessary due to the length of the input data.

obgm (commented):
In the last sentence of the commit message: s/unsisgned/unsigned/ (or, size_t, respectively)

ecc/ecc.c
obgm: Ah, sorry, I did not realize that this has already been fixed in #81.

benpicco:
Let's close this in favor of #81

PR #83, opened 2021-06-04
tests: Handle IPv4/IPv6 selection better

mrdeep1:
tests/dtls-client.c:

Fix ALERT generated when communicating with an IPv4 host.

tests/dtls-server.c:

Support binding to an IPv4 address.


obgm:
I do not exactly remember what IPv4 was. Anyway...

PR #86, opened 2021-06-07
dtls.c: Update logging for received records

mrdeep1:
Record epoch as well as sequence number.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -688,7 +688,8 @@ static void dtls_debug_keyblock(dtls_security_parameters_t *config)
   * see IANA for a full list of types:
   * https://www.iana.org/assignments/tls-parameters/tls-parameters.xml#tls-parameters-7
   */
-static const char *dtls_handshake_type_to_name(int type)
+static const char *
+dtls_handshake_type_to_name(int type)
 {
   switch (type) {
   case DTLS_HT_HELLO_REQUEST:
@@ -717,6 +718,24 @@ static const char *dtls_handshake_type_to_name(int type)
     return "unknown";
   }
 }
+
+static const char *
+dtls_message_type_to_name(int type)
+{
+  switch (type) {
+  case DTLS_CT_CHANGE_CIPHER_SPEC:
+    return "change_cipher_spec";
+  case DTLS_CT_ALERT:
+    return "alert";
+  case DTLS_CT_HANDSHAKE:
+    return "handshake";
+  case DTLS_CT_APPLICATION_DATA:
+    return "application_data";
+  default:
+    return NULL;
+  }
+}
+
 
 /**
  * Calculate the pre master secret and after that calculate the master-secret.
@@ -3756,7 +3775,8 @@ handle_handshake(dtls_context_t *ctx, dtls_peer_t *peer, session_t *session,
   hs_header = DTLS_HANDSHAKE_HEADER(data);
 
   dtls_debug("received handshake packet of type: %s (%i)\n",
-	     dtls_handshake_type_to_name(hs_header->msg_type), hs_header->msg_type);
+             dtls_handshake_type_to_name(hs_header->msg_type),
+             hs_header->msg_type);
 
   packet_length = dtls_uint24_to_int(hs_header->length);
   fragment_length = dtls_uint24_to_int(hs_header->fragment_length);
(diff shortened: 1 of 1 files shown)


dtls.c
```diff
@@ -714,10 +715,30 @@ static const char *dtls_handshake_type_to_name(int type)
   case DTLS_HT_FINISHED:
     return "finished";
   default:
-    return "unknown";
+    snprintf(buf, sizeof(buf), "unknown (%d)", type);
```
obgm, line 718: I would like to avoid `snprintf()` in tinydtls as it is not available on many non-POSIX embedded platforms. As this is used only for converting the handshake type number to ASCII characters, we could use a small helper function instead, e.g.:

```c
static size_t print_byte(char *buf, size_t buflen, uint8_t value) {
  char const *bufstart = buf;
  uint8_t quot = 100;

  if (value == 0) {
    *buf++ = '0';
    buflen--;
  } else {
    /* adjust quot to avoid leading zeroes */
    while (quot && (value / quot == 0))
      quot /= 10;

    while (quot && buflen) {
      *buf++ = '0' + (value / quot);
      buflen--;
      value %= quot;
      quot /= 10;
    }
  }
  return buf - bufstart;
}
```

Usage would be:

```c
  static char buf[16] = "unknown (";
  size_t pos = 9;

  pos += print_byte(buf + pos, sizeof(buf) - pos - 2, type & 0xff);
  assert(pos+2 < sizeof(buf));
  buf[pos++] = ')';
  buf[pos] = '\000';
```

dtls.c
```diff
@@ -714,10 +715,30 @@ static const char *dtls_handshake_type_to_name(int type)
...
+  case DTLS_CT_CHANGE_CIPHER_SPEC:
+    return "change_cipher_spec";
+  case DTLS_CT_ALERT:
+    return "alert";
+  case DTLS_CT_HANDSHAKE:
+    return "handshake";
+  case DTLS_CT_APPLICATION_DATA:
+    return "application_data";
+  default:
+    snprintf(buf, sizeof(buf), "unknown (%d)", type);
```
obgm, line 736: See previous comment.

dtls.c
boaks: A hexdec value may do it as well ... and may be easier to `print`.

mrdeep1: Taken code suggestion.

dtls.c
mrdeep1: Updated as per previous comment and code pushed.

dtls.c
mrdeep1: Just seen this

> A hexdec value may do it as well ... and may be easier to print.

Not familiar with hexdec, but I think I understand the principle.  `print_byte()` could be updated I guess.

obgm: I thought about printing hex values but went for decimal as the handshake types in the RFC are specified in decimal.

dtls.c
```diff
@@ -684,12 +684,37 @@ static void dtls_debug_keyblock(dtls_security_parameters_t *config)
...
+  return buf - bufstart;
+}
+
 /** returns the name of the goven handshake type number.
   * see IANA for a full list of types:
   * https://www.iana.org/assignments/tls-parameters/tls-parameters.xml#tls-parameters-7
   */
 static const char *dtls_handshake_type_to_name(int type)
 {
+ static char buf[16] = "unknown (";;
```
boaks, line 716: Code as:

```
#ifdef DTLS_CONSTRAINED_STACK
  dtls_mutex_lock(&static_mutex);
#endif /* DTLS_CONSTRAINED_STACK */
```

makes me wonder, if concurrent processing is assumed?

dtls.c
```diff
@@ -714,10 +739,37 @@ static const char *dtls_handshake_type_to_name(int type)
...
-    return "unknown";
+    pos += print_byte(buf + pos, sizeof(buf) - pos - 2, type & 0xff);
+    assert(pos+2 < sizeof(buf));
+    buf[pos++] = ')';
+    buf[pos] = '\000';
+    return buf;
+  }
+}
+
+static const char *dtls_message_type_to_name(int type)
```
boaks, line 750: whats about returning NULL and use two dtls_debug?

```
tyepAsText=dtls_message_type_to_name(msg[0]);
if (typeAsText) {
   dtls_debug("got '%s' epoch %u sequence %" PRIu64 " (%d bytes)\n",
               typeAsText, dtls_get_epoch(header), pkt_seq_nr, rlen);
} else {
   dtls_debug("got unknown record type (%d) epoch %u sequence %" PRIu64 " (%d bytes)\n",
               msg[0], dtls_get_epoch(header), pkt_seq_nr, rlen);
}
```

mrdeep1: Gone with the principal of this suggestion and done some other tidy up. Code pushed.

dtls.c
mrdeep1: Interesting question - multi-threading is not supported as far as I am aware.  Looks like I added this in some time back to remove data held on the stack into a static area if the stack was constrained in size and wanted to make sure things were properly protected.

obgm:
I like this. Can you please squash the commits for merge?

mrdeep1:
Commits quashed and pushed.

PR #87, opened 2021-06-07
dtls.c: Destroy peer on generation of fatal alert during handshake

mrdeep1:
RFC 5246 7.2 Alert Protocol

   Alert messages
   with a level of fatal result in the immediate termination of the
   connection.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -4186,7 +4186,15 @@ dtls_handle_message(dtls_context_t *ctx,
       if (err < 0) {
 	dtls_warn("error while handling handshake packet\n");
 	dtls_alert_send_from_err(ctx, peer, session, err);
-	return err;
+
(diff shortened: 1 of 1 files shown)


dtls.c
```diff
@@ -4122,7 +4122,17 @@ dtls_handle_message(dtls_context_t *ctx,
       if (err < 0) {
 	dtls_warn("error while handling handshake packet\n");
 	dtls_alert_send_from_err(ctx, peer, session, err);
-	return err;
+
+        if (peer && err < -(1 << 8) && err > -(3 << 8)) {
+          if (DTLS_ALERT_LEVEL_FATAL == ((-err) & 0xff00) >> 8) {
```
obgm, line 4127: It seems to me that this check is essentially the same as the check for `err` in the previous line. As this one is more explicit in what is really being checked, I suggest keeping this one and get rid of the other.

mrdeep1: Updated and code pushed.

PR #88, opened 2021-06-07
dtls.c: Check for duplicate optional handshake messages

mrdeep1:
In general the handshake state machine forces the handshakes to go through
a defined order, but duplicate optional CertificateRequest or ServerKeyExchange
handshakes are not detected.

This fix adds in tracking whether a CertificateRequest or ServerKeyExchange
have been received and rejects any duplication.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -3291,6 +3291,8 @@ dtls_renegotiate(dtls_context_t *ctx, const session_t *dst)
 
   peer->handshake_params->hs_state.mseq_r = 0;
   peer->handshake_params->hs_state.mseq_s = 0;
+  peer->serverkeyexchange = 0;
+  peer->certificaterequest = 0;
 
   if (peer->role == DTLS_CLIENT) {
     /* send ClientHello with empty Cookie */
@@ -3400,7 +3402,8 @@ handle_handshake_msg(dtls_context_t *ctx, dtls_peer_t *peer, session_t *session,
 #endif /* DTLS_ECC */
 #ifdef DTLS_PSK
     if (is_tls_psk_with_aes_128_ccm_8(peer->handshake_params->cipher)) {
-      if (state != DTLS_STATE_WAIT_SERVERHELLODONE) {
+      if (state != DTLS_STATE_WAIT_SERVERHELLODONE ||
+          peer->serverkeyexchange || peer->certificaterequest) {
         return dtls_alert_fatal_create(DTLS_ALERT_UNEXPECTED_MESSAGE);
       }
       err = check_server_key_exchange_psk(ctx, peer, data, data_length);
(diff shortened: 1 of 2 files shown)


boaks:
This PR solves the double processing of the two messages. 
But a reorder will not be detected (certificate request before server key exchange).

mrdeep1:
handle_message() should do the necessary buffering to handle any packet arrival re-ordering before calling handle_handshake().

Agree that a CertificateRequest cannot be before ServerKeyExchange.  Updated code pushed.

boaks:
It was not about "arriving" reordered, it's more about send by intention in a wrong order (including a wrong order in the message sequence number).

dtls.c
```diff
@@ -3400,7 +3402,8 @@ handle_handshake_msg(dtls_context_t *ctx, dtls_peer_t *peer, session_t *session,
 #endif /* DTLS_ECC */
 #ifdef DTLS_PSK
     if (is_tls_psk_with_aes_128_ccm_8(peer->handshake_params->cipher)) {
-      if (state != DTLS_STATE_WAIT_SERVERHELLODONE) {
+      if (state != DTLS_STATE_WAIT_SERVERHELLODONE ||
+          peer->serverkeyexchange || peer->certificaterequest) {
```
boaks, line 3406: `peer->certificaterequest ` should never be 1 for PSK, or?

boaks:
I found, that my latest comments are in contradiction to my firsts.

As far as I understand the implementation now, the state machines for ECDHE_ECDSA and PSK are (obviously) different:

ECDHE_ECDSA:

- DTLS_STATE_WAIT_SERVERCERTIFICATE => accept Certificate      
- DTLS_STATE_WAIT_SERVERKEYEXCHANGE => accept ServerKeyExchange
- DTLS_STATE_WAIT_SERVERHELLODONE => accept (optional) CertificateRequest, accept ServerHelloDone
- DTLS_STATE_WAIT_CHANGECIPHERSPEC

PSK: 

- DTLS_STATE_WAIT_SERVERHELLODONE => accept (optional) ServerKeyExchange, accept ServerHelloDone


With that: my comment above

> This PR solves the double processing of the two messages.
> But a reorder will not be detected (certificate request before server key exchange).
> 
> Adding one more || peer->certificaterequest to if (state != DTLS_STATE_WAIT_SERVERHELLODONE || peer->serverkeyexchange) should fix that as well.

is wrong! 
For PSK a CertificateRequest is always wrong, and the sequence related to ServerKeyExchange doesn't apply, because it would have already been an error. Therefore using peer->serverkeyexchange is sufficient.

For ECDHE_ECDSA  ServerKeyExchange is mandatory and checked by its state. The optional CertificateRequest is then protected by  peer->certificaterequest.

boaks:
Hi  @mrdeep1 ,

is it still on the way to the top? 
Just, if you're too short with time and you would prefer, that I prepare an alternative PR, that would be also a way to go.

mrdeep1:
@boaks there are too many other things going on at present to give this my full attention.  If you have the time, it would be appreciated if you can do the work - thanks.

boaks:
@mrdeep1 

With merging the alternative approach in PR #109, this may be closed?

mrdeep1:
Yes, PR #109 looks to be doing the same thing (and better) from doing a code inspection - thanks.

PR #89, opened 2021-06-09
Add separate stateless processing for client_hello of epoch 0.

boaks:
The PR is on the top of some other pending ones.
If at all merged, it is intended to be merged after these PRs.

Use the record and message sequence numbers of that client_hellos for
either sending a hello verify request or starting the handshake on the
server side with sending a server_hello.

The function of a HELLO_REQUEST is unclear and therefore not considered.
In some scopes a HELLO_REQUEST is only used for associated peers to
trigger a rehandshake (handshake executed within an previous encrypted
association, epoch > 0), and some scopes seems to handle also
situations, where either no previous association is required or the new
handshake is executed in epoch 0. That may be clarified either before or
after this change.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -123,6 +123,8 @@ memarray_t dtlscontext_storage;
 #define DTLS_CV_LENGTH (1 + 1 + 2 + 1 + 1 + 1 + 1 + DTLS_EC_KEY_SIZE + 1 + 1 + DTLS_EC_KEY_SIZE)
 #define DTLS_FIN_LENGTH 12
 
+#define DTLS_ALERT_LENGTH 2 /* length of the Alert message */
+
 #define HS_HDR_LENGTH  DTLS_RH_LENGTH + DTLS_HS_LENGTH
 #define HV_HDR_LENGTH  HS_HDR_LENGTH + DTLS_HV_LENGTH
 
@@ -471,12 +473,14 @@ is_record(uint8 *msg, size_t msglen) {
 /**
  * Initializes \p buf as record header. The caller must ensure that \p
  * buf is capable of holding at least \c sizeof(dtls_record_header_t)
- * bytes. Increments sequence number counter of \p security.
+ * bytes. Increments records sequence number counter.
  * \return pointer to the next byte after the written header.
  * The length will be set to 0 and has to be changed before sending.
  */
 static inline uint8 *
-dtls_set_record_header(uint8 type, dtls_security_parameters_t *security,
+dtls_set_record_header(uint8 type,
+		       uint16_t epoch,
+		       uint64_t *rseqn,
 		       uint8 *buf) {
   dtls_int_to_uint8(buf, type);
   buf += sizeof(uint8);
@@ -484,20 +488,16 @@ dtls_set_record_header(uint8 type, dtls_security_parameters_t *security,
   dtls_int_to_uint16(buf, DTLS_VERSION);
   buf += sizeof(uint16);
 
-  if (security) {
-    dtls_int_to_uint16(buf, security->epoch);
-    buf += sizeof(uint16);
+  dtls_int_to_uint16(buf, epoch);
+  buf += sizeof(uint16);
 
-    dtls_int_to_uint48(buf, security->rseq);
-    buf += sizeof(uint48);
+  dtls_int_to_uint48(buf, *rseqn);
+  buf += sizeof(uint48);
 
-    /* increment record sequence counter by 1 */
-    security->rseq++;
-  } else {
-    memset(buf, 0, sizeof(uint16) + sizeof(uint48));
-    buf += sizeof(uint16) + sizeof(uint48);
-  }
+  /* increment record sequence counter by 1 */
+  (*rseqn)++;
 
+  /* space for record size */
   memset(buf, 0, sizeof(uint16));
   return buf + sizeof(uint16);
 }
@@ -505,11 +505,12 @@ dtls_set_record_header(uint8 type, dtls_security_parameters_t *security,
 /**
  * Initializes \p buf as handshake header. The caller must ensure that \p
  * buf is capable of holding at least \c sizeof(dtls_handshake_header_t)
- * bytes. Increments message sequence number counter of \p peer.
+ * bytes. Increments message sequence number counter.
  * \return pointer to the next byte after \p buf
  */
 static inline uint8 *
-dtls_set_handshake_header(uint8 type, dtls_peer_t *peer,
+dtls_set_handshake_header(uint8 type,
+			  uint16_t *mseqn,
 			  int length,
 			  int frag_offset, int frag_length,
 			  uint8 *buf) {
@@ -520,16 +521,12 @@ dtls_set_handshake_header(uint8 type, dtls_peer_t *peer,
   dtls_int_to_uint24(buf, length);
   buf += sizeof(uint24);
 
-  if (peer && peer->handshake_params) {
-    /* and copy the result to buf */
-    dtls_int_to_uint16(buf, peer->handshake_params->hs_state.mseq_s);
-
-    /* increment handshake message sequence counter by 1 */
-    peer->handshake_params->hs_state.mseq_s++;
-  } else {
-    memset(buf, 0, sizeof(uint16));
-  }
+  /* and copy the result to buf */
+  dtls_int_to_uint16(buf, *mseqn);
   buf += sizeof(uint16);
+
+  /* increment handshake message sequence counter by 1 */
+  (*mseqn)++;
 
   dtls_int_to_uint24(buf, frag_offset);
   buf += sizeof(uint24);
@@ -624,38 +621,7 @@ known_cipher(dtls_context_t *ctx, dtls_cipher_t code, int is_client) {
 	 (ecdsa && is_tls_ecdhe_ecdsa_with_aes_128_ccm_8(code));
 }
 
-/**
- * This method detects if we already have a established DTLS session with
- * peer and the peer is attempting to perform a fresh handshake by sending
- * messages with epoch = 0. This is to handle situations mentioned in
- * RFC 6347 - section 4.2.8.
- *
- * @param msg  The packet received from Client
- * @param msglen Packet length
- * @param peer peer who is the sender for this packet
- * @return @c 1 if this is a rehandshake attempt by
- * client
- */
-static int
-hs_attempt_with_existing_peer(uint8_t *msg, size_t msglen,
-    dtls_peer_t *peer)
-{
-    (void) msglen;
-    if ((peer) && (peer->state == DTLS_STATE_CONNECTED)) {
-      if (msg[0] == DTLS_CT_HANDSHAKE) {
-        uint16_t msg_epoch = dtls_uint16_to_int(DTLS_RECORD_HEADER(msg)->epoch);
-        if (msg_epoch == 0) {
-          dtls_handshake_header_t * hs_header = DTLS_HANDSHAKE_HEADER(msg + DTLS_RH_LENGTH);
-          if (hs_header->msg_type == DTLS_HT_CLIENT_HELLO ||
-              hs_header->msg_type == DTLS_HT_HELLO_REQUEST) {
-            return 1;
-          }
-        }
-      }
-    }
-    return 0;
-}
-/** Dump out the cipher keys and IVs used for the symetric cipher. */
+/** Dump out the cipher keys and IVs used for the symmetric cipher. */
 static void dtls_debug_keyblock(dtls_security_parameters_t *config)
 {
   dtls_debug("key_block (%d bytes):\n", dtls_kb_size(config, peer->role));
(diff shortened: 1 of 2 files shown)


boaks:
Together with the other PRs #91, #92, #94, this PR provides now a redesign and cleanup for processing CLIENT_HELLO in epoch 0 and removes some stuff, which seems for me to be work-arounds.

I removed the rudiments of the HELLO_REQUEST in epoch 0.
If really someone complains, it may be introduced again, then with details on how that should be processed.

obgm:
Thanks. May I ask you to change the spelling of DTLS messages in the commit message and in code comments in alignment with RFC 6347 (i.e., camel case: ClientHello, HelloRequest, Finished, etc.)?

boaks:
Sure. I'm on it.

boaks:
Done.

Do you prefer that CaMel also in the c-documentation?

obgm:
Yes, please.

boaks:
The comments should be fixed.

Let me know, if there is more editorial work to do. I will be back in the later afternoon :-).

mrdeep1:
If I try using a bad PSK, the server fails to decrypt the Finished, which is fine.  As there is no response to the 'bad' Finish the client retries sending the stuff from epoch 0 and then the Finished again in epoch 1.

The previous code reports
````
WARN Wrong epoch, expected 1, got: 0
WARN Wrong epoch, expected 1, got: 0
WARN Wrong epoch, expected 1, got: 0
````
and the new code
````
WARN The message sequence number is too small, expected 4, got: 1
WARN The message sequence number is too small, expected 4, got: 2
WARN The message sequence number is too small, expected 4, got: 3
````
As the server has done a ChangeCipherSpec and moved into epoch 1, should the server be continuing with the stateless processing for epoch 0?

As the Finished cannot be decrypted, there is no way of determining what the actual handshake is, but we know that it is in the new epoch as a handshake with a small record sequence number.  Should we just drop this 'bad' record (as does OpenSSL) or send back an alert (as does GnuTLS - Bad Record MAC)?

**Comment by boaks** [comment]
I'm not sure, if I understand the scenario.

The "stateless" end with a ClientHello with a verified cookie. 
After that, there is a peer.
Is it possible to see the wireshark capture?

**Comment by mrdeep1** [comment]
Sure - abstracted an example out of my test suite which gives slightly different results to what I reported above. 
[pr_89-1.pcap.zip](https://github.com/eclipse/tinydtls/files/6735333/pr_89-1.pcap.zip)

Server key was 12345678 , client key 87654321

Coap-Server log
````
Jun 29 18:35:21 WARN decryption failed
Jun 29 18:35:23 WARN The message sequence number is too small, expected 3, got: 2
Jun 29 18:35:23 WARN expected ChangeCipherSpec during handshake
Jun 29 18:35:23 WARN decryption failed
Jun 29 18:35:27 WARN The message sequence number is too small, expected 3, got: 2
Jun 29 18:35:27 WARN expected ChangeCipherSpec during handshake
Jun 29 18:35:27 WARN decryption failed
Jun 29 18:35:35 WARN The message sequence number is too small, expected 3, got: 2
Jun 29 18:35:35 WARN expected ChangeCipherSpec during handshake
Jun 29 18:35:35 WARN decryption failed
Jun 29 18:35:46 WARN decryption failed
Jun 29 18:35:59 WARN decryption failed
Jun 29 18:36:01 WARN The message sequence number is too small, expected 3, got: 2
Jun 29 18:36:01 WARN expected ChangeCipherSpec during handshake
Jun 29 18:36:01 WARN decryption failed
Jun 29 18:36:05 WARN The message sequence number is too small, expected 3, got: 2
Jun 29 18:36:05 WARN expected ChangeCipherSpec during handshake
Jun 29 18:36:05 WARN decryption failed
Jun 29 18:36:13 WARN The message sequence number is too small, expected 3, got: 2
Jun 29 18:36:13 WARN expected ChangeCipherSpec during handshake
Jun 29 18:36:13 WARN decryption failed
Jun 29 18:36:19 WARN decryption failed
````

> The "stateless" end with a ClientHello with a verified cookie.
> After that, there is a peer.

Agreed, however the code for `case DTLS_CT_HANDSHAKE:` in `dtls_handle_message()` no longer checks that the handshake is in the right epoch and so goes on the report the `The message sequence number is too small, expected `.

**Comment by boaks** [comment]
> As the Finished cannot be decrypted, there is no way of determining what the actual handshake is, but we know that it is in the new epoch as a handshake with a small record sequence number. Should we just drop this 'bad' record (as does OpenSSL) or send back an alert (as does GnuTLS - Bad Record MAC)?

I'm not sure, if this is intended. At least for Californium there was a discussion about that [PSK-failures](https://github.com/eclipse/californium/issues/606) and in that context [TLS-IETF](https://mailarchive.ietf.org/arch/msg/tls/sbPuZCw5zOs2LpkAPX6I2isGBTM/).

**Comment by boaks** [comment]
> my test suite 

That's not in the tinydtls repo, or?
(I was already asking me, if there are more tests ...).

**Comment by mrdeep1** [comment]
No, it is not in the tinydtls repo.

I have built up a regression test suite for libcoap over time for testing out different functionality and using different (D)TLS libraries with good/bad keys/certs etc. Diffs are then done for actual against expected.  When using valgrind, things grind on for hours.....

Unfortunately, it is very dependent on the host OS, versions of underlying (D)TLS libraries that were would be too many false positives if it was to became part of the libcoap repo.

**Comment by boaks** [comment]
> Unfortunately, it is very dependent on the host OS, versions of underlying (D)TLS libraries that were would be too many false positives if it was to became part of the libcoap repo.

OK. Good to know, that there are more tests.

**Comment by boaks** [comment]
I checked the GnuTLS behavior, it closes on the retransmission of flight 5. So the first flight 5 seems to be  ignored.

Even in the Californium discussion on 2018, I wasn't totally convinced , what the best behavior will be. Anyway, usually to systems are used with the "right credentials", so I don't care too much about the wrong ;-).

**Comment by boaks** [comment]
My feeling is, "alert or not", is not really changed by this PR (hope I'm right.)
But by other PRs before, I guess #91 .

**Comment by mrdeep1** [comment]
Prior to #91, Decrypt Error is returned when processing the Finished.  This PR (which includes #91) changes the message that is reported.

RFC5246 says
`````
   decrypt_error
      A handshake cryptographic operation failed, including being unable
      to correctly verify a signature or validate a Finished message.
      This message is always fatal.
````
But agree that RFC6347 effectively states that Alerts should not be sent.

I am not sure that `check_finished()` will ever get called if the wrong keys are used as it is not possible to decrypt the Finished and so there is no way of working out the handshake type (other than the clue of epoch change and small sequence no).

**Comment by boaks** [comment]
> I am not sure that `check_finished()` will ever get called 

That protects against "manipulated handshake messages", e.g. on-path-attacker.
If extended master secret is use, this will happen really rarely.

**Comment by boaks** [comment]
And using RPK changes the possible failure as well. There the encrytion may work more often, but manipulated handshake message may get detected.

**Comment by boaks** [comment]
> Yes, please.

The c-documentation is updated to use CaMeL. I added some more information about the "stateless phase" (which ends with receiving a ClientHello with a valid cookie). I adapted the function names of the affected funtions to a pattern xxx_0_yyy, in order to easier detect them (the other difference is the ephemeral_peer instead of a peer).

**Comment by obgm** [comment]
@boaks Thank you, this looks good to me. I am not sure if the two commits should be squashed before merge. Do you have an opinion on that?

**Comment by boaks** [comment]
> I am not sure if the two commits should be squashed before merge.

I would do so. 

Though this PR together with #91 obsoletes #85, @mrdeep1 is that also OK for you?

**Comment by mrdeep1** [comment]
Fine by me.

**Comment by boaks** [comment]
OK, I squash and push.

File: dtls.c [hunk 1]
```diff
@@ -1375,25 +1341,25 @@ check_finished(dtls_context_t *ctx, dtls_peer_t *peer,
...
- * \param rlen    This parameter must be initialized with the
- *                maximum size of \p sendbuf and will be updated
- *                to hold the actual size of the stored packet
- *                on success. On error, the value of \p rlen is
- *                undefined.
+ * \param peer            The remote peer the packet will be sent to.
+ * \param security        The encryption paramater used to encrypt
+ * \param type            The content type of this record.
+ * \param data_array      Array with payloads in correct order.
+ * \param data_len_       array sizes of the payloads in correct order.
```
obgm, line 1354: Whitespace error: `data_len_array`

dtls.c
```diff
@@ -1539,6 +1510,98 @@ dtls_prepare_record(dtls_peer_t *peer, dtls_security_parameters_t *security,
   return 0;
 }
 
+/**
+ * Send Alert in stateless fashion.
+ * A Alert is sent to the peer (using the write callback function
```
obgm, line 1515: "A Alert" → "An Alert message"

dtls.c
```diff
@@ -1539,6 +1510,98 @@ dtls_prepare_record(dtls_peer_t *peer, dtls_security_parameters_t *security,
...
+ * registered with \p ctx). The return value is \c -1 on error, \c 1 when
+ * the Alert is sent.
+ *
+ * \param ctx              The DTLS context.
+ * \param ephemeral_peer   The ephemeral remote party we are talking to.
+ * \param level            Alert level.
+ * \param description      Alert description.
+ * \return \c 1 if msg is sent, or \c -1 otherwise.
```
obgm, line 1523: Actually, the result is "a value less than zero" on error (as stated in the documentation of the `write` callback for `dtls_handler_t`) or "the number of bytes sent".

dtls.c
```diff
@@ -1539,6 +1510,98 @@ dtls_prepare_record(dtls_peer_t *peer, dtls_security_parameters_t *security,
...
+
+  dtls_debug_hexdump("send header", buf, DTLS_RH_LENGTH);
+  dtls_debug_hexdump("send unencrypted alert", p, DTLS_ALERT_LENGTH);
+
+  int res = CALL(ctx, write, ephemeral_peer->session, buf, sizeof(buf));
+
+  /* Guess number of bytes application data actually sent:
+   * dtls_prepare_record() tells us in len the number of bytes to
+   * send, res will contain the bytes actually sent. */
+  return res <= 0 ? res : (int) sizeof(buf);
```
obgm, line 1551: It is unclear to me why `res` is not passed through, especially since `sizeof(buf)` ≠ 1 (c.f. doxygen comment for `\return`).

dtls.c
```diff
@@ -1539,6 +1510,98 @@ dtls_prepare_record(dtls_peer_t *peer, dtls_security_parameters_t *security,
...
+  dtls_debug_hexdump("send header", buf, DTLS_RH_LENGTH);
+  dtls_debug_hexdump("send unencrypted handshake header", buf + DTLS_RH_LENGTH, DTLS_HS_LENGTH);
+  dtls_debug_hexdump("send unencrypted cookie", data, data_length);
+
+  res = CALL(ctx, write, ephemeral_peer->session, buf, sizeof(buf));
+
+  /* Guess number of bytes application data actually sent:
+   * dtls_prepare_record() tells us in len the number of bytes to
+   * send, res will contain the bytes actually sent. */
+  return res <= 0 ? res : (int) sizeof(buf);
```
obgm, line 1602: Here, the same holds as for the doxygen comment and return statement of `dtls_0_send_alert()`.

dtls.c
```diff
@@ -4031,19 +4142,61 @@ dtls_handle_message(dtls_context_t *ctx,
 				   (without MAC and padding) */
   int err;
 
+  /* check for CLIENT_HELLOs of epoch 0, maybe a peer's start over */
```
obgm, line 4145: "CLIENT_HELLOs" → "ClientHellos"

dtls.c
```diff
@@ -4031,19 +4142,61 @@ dtls_handle_message(dtls_context_t *ctx,
...
+    return 0;
+  }
+
   /* check if we have DTLS state for addr/port/ifindex */
   peer = dtls_get_peer(ctx, session);
 
   if (!peer) {
     dtls_debug("dtls_handle_message: PEER NOT FOUND\n");
     dtls_dsrv_log_addr(DTLS_LOG_DEBUG, "peer addr", session);
+    /** no peer, no CLIENT_HELLO => drop it */
```
obgm, line 4193: "CLIENT_HELLO" →"ClientHello"

obgm (commented):
A few minor nits, otherwise I think it is ready for merge.

dtls.c
boaks: OK

dtls.c
boaks: OK

dtls.c
boaks: Ok

dtls.c
boaks: Copied from other functions :-). 
Replaced by res.

dtls.c
boaks: OK

dtls.c
boaks: ok

dtls.c
boaks: ok

PR #94, opened 2021-06-26
Fix wrong read epoch

boaks:


dtls.c
```diff
@@ -4026,6 +4029,7 @@ handle_ccs(dtls_context_t *ctx, dtls_peer_t *peer,
     }
   }
 
+  peer->handshake_params->hs_state.read_epoch++;
```
obgm, line 4032: We might want to add `assert(... read_epoch) > 0` below this line as a reminder that at some point, overly large epochs need to be addressed.

peer.h
```diff
@@ -76,6 +76,18 @@ static inline dtls_security_parameters_t *dtls_security_params_epoch(dtls_peer_t
   }
 }
 
+static inline dtls_security_parameters_t *dtls_security_params_read_epoch(dtls_peer_t *peer, uint16_t epoch)
```
obgm, line 79: I suggest a comment that explains this function's behavior as it requires at least careful reading.

boaks: OK.

dtls.c
boaks: OK.

obgm (approved):

Issue #95, opened 2021-07-02
DTLS_STATE_CLOSING may block session slots


obgm:
RIOT-OS/RIOT#16422 correctly states that calling `dtls_close()` may result in a session held in state `DTLS_STATE_CLOSING` forever if no alert messages from the peer are received. Reasons for this might be dropped packets or a crashed/gone peer. A viable solution could include triggering cleanup (either through `dtls_reset_peer()` or `dtls_destroy_peer()`) after some timeout.

boaks:
[RFC5246 - 7.2.1.  Closure Alerts](https://datatracker.ietf.org/doc/html/rfc5246#section-7.2.1)

> It is not required for the initiator of the close to wait for the responding close_notify alert before
> closing the read side of the connection.

So, yes, that wait in DTLS_STATE_CLOSING may time out.

In my personal opinion, there are no good reason to use a Closure Alerts for DTLS at all.
The most definitions of RFC5246 (e.g. "required") will require to ACK/retransmit the Closure Alerts for RFC6347 in order to ensure, that the other peer received it.

[RFC6347 - 4.2.7.  Alert Messages](https://datatracker.ietf.org/doc/html/rfc6347#section-4.2.7)

> Note that Alert messages are not retransmitted at all

So, the means to ensure, that a "require"  Closure Alerts is received by the other side, are not used.

- truncation attack
- continue to communicate without encrpytion
-  "It is assumed that closing a connection reliably delivers
   pending data before destroying the transport."

Therefore all that hardly applies to DTLS without ack/retransmission, because without them, Closure Alerts are reduced to "may be received".

obgm:
I think that closure alerts are still a good thing as an entity can react immediately on receipt of close_notify while it would have to rely on timeouts otherwise. Thus, these one-shot alerts are useful in the best case. The trade-off taken was apparently between stale connections and truncation attacks.

boaks:
I guess, the question is the use-pattern.

If DTLS is used in a "time boxed session" way, then the Closure Alerts and new (resumption) handshakes for the next time makes sense. (With resumption, only the association keys could be freed, the master secret is still kept for resumption. So the benefit of saving resources may not be that large.)

If DTLS is used for "mid-term communication" in order to have less state for communication as CoAP may offer, then handshakes are used e.g once a day or week, and encrypted messages are exchanged without a prior handshake. That's mainly a question of the expected communication behavior of the application.

mrdeep1:
[...]

PR #96, opened 2021-07-10
Fix Alert timeout

boaks:
Fix waiting without timeout in DTLS_STATE_CLOSING after Alert.

Sending Alert according RFC6347 without retransmission is not reliable.
Therefore a dummy retransmission entry is used for outgoing Alerts. On
retransmission, the Alert is not retransmitted, instead handle_alert is
called to emulate receiving the response.

Issue #95

### DIFF
--- a/netq.h
+++ b/netq.h
@@ -39,6 +39,14 @@
 #endif
 #endif
 
+/**
+ * Job to be executed for the timer entry in the netq.
+ */
+typedef enum netq_job_type_t {
+  RESEND, 	/**< resend related message on timeout */
+  TIMEOUT 	/**< timeout of the related alert */
+} netq_job_type_t;
+
 /** 
  * Datagrams in the netq_t structure have a fixed maximum size of
  * DTLS_MAX_BUF to simplify memory management on constrained nodes. */ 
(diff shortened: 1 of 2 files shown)


boaks:
Just to mention:

Using this with the libcoap examples results in leaks reported by valgrind.
I'm not sure, if the "cleanup" must take care of such retransmission jobs, or if the "shutdown" must now wait for the response or timeout.

boaks:
One idea to fix the resulting leak is to add a `dtls_stop_retransmission` call to `dtls_destroy_peer`. For me this makes sense, but I'm not sure, if the use-cases, which this is not intended. If a `dtls_stop_retransmission` call is added to `dtls_destroy_peer`, then the most caller of `dtls_destroy_peer` may remove it ... which would obsolete the extra `dtls_reset_peer`, though that adds only the obsoleted `dtls_stop_retransmission`.

netq.h
```diff
@@ -39,6 +39,10 @@
 #endif
 #endif
 
+typedef enum netq_job_type_t {
```
obgm, line 42: The new `netq_job_type` and the new field `job` in `netq_t` should be documented.

obgm (commented):
Looks good except for the missing documentation of the new enum and field in netq.h, see below.

netq.h
boaks: OK. I will do it a little later in the afternoon.

boaks:
I added some documentation. Not sure, if you prefer to have more details on the jobs and how the processing will be done. If so, let me know.

PR #101, opened 2021-08-07
Fix issue 97

### LINKED ISSUE #97: Infinite Loop Bug Affecting handling ClientHello Handkshake message
Hello,

In the project master branch, an infinite loop bug had been found during handling ClientHello Handshake message. Suppose you have a malicious ClientHello message with the following values for the mentioned fields:
    1. **Cipher Suites Length** is an odd number like 3 (this is a malicious number)
    2. **Two Suites** whose length is 4 bytes
    3. **Extension Length** whose length is not zero like 26 bytes

After handling this message in the normal handling way as follows, we enter into the function **dtls_update_parameters()** at the dtls.c:980.
    1. 0x4e2b02 in dtls_update_parameters /home/etinydtls/dtls.c:1024:22
    2. 0x4e2b02 in handle_handshake_msg /home/etinydtls/dtls.c:3403:11
    3. 0x4cb1ce in handle_handshake /home/etinydtls/dtls.c:3493:14
    4. 0x4cb1ce in dtls_handle_message /home/etinydtls/dtls.c:3881:13
    5. 0x4c4efc in dtls_handle_read /home/etinydtls/tests/dtls-server.c:177:10
    6. 0x4c4efc in main /home/etinydtls/tests/dtls-server.c:352:2
    7. 0x7fb176ab10b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
    8. 0x41c45d in _start (/home/etinydtls/tests/dtls-server+0x41c45d)

When coming to dtls.c:1005 in the function dtls_update_parameters(), variable **i** is assigned to a int number, which stands for all cipher suites' length 3 in this message. After that, the condition **data_length < i + sizeof(uint16)** cannot refuse this packet cannot exclude this packet, as 4+26 < 3+2 is no satisfied. Hence, we come to the loop of confirming suites at the dtls.c:1023, where **i** is decreased by 2 each loop. The ending loop condition includes (i == 0), while the initial value of i is 3 thereby this condition will not be satisfied. Unfortunately, one infinite loop incurs, which lets the server loop forever until consuming all resources. 

Here is a message packet to reproduce this bug: [https://github.com/jerrytesting/dtls-infiniteLoop.git]

I hope I could explain it clearly. Could you kindly have a check? Thanks a lot.

Kind Regards,
Jerry

boaks:
Fixes #97 

Fail on odd cipher suite length.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -1118,6 +1118,11 @@ dtls_update_parameters(dtls_context_t *ctx,
     return 0;
   }
 
+  if ((i % sizeof(uint16)) != 0) {
+    dtls_debug("odd length for cipher suites\n");
+    goto error;
+  }
+
   data += sizeof(uint16);
   data_length -= sizeof(uint16) + i;
 


dtls.c
```diff
@@ -1163,9 +1168,6 @@ dtls_update_parameters(dtls_context_t *ctx,
     data += sizeof(uint16);
   }
 
-  /* skip remaining ciphers */
-  data += i;
```
obgm, line 1167: I think that this is necessary here as `data` needs to point to the start of the compression methods below.

obgm (commented):
May I ask you to state the nature of the bugfix in the first line of the commit message?
The commit message then may include "Fixes: issue 97".

dtls.c
boaks: Yes, I overseen the "ok" case.

obgm (approved):

PR #103, opened 2021-10-09
fix build with automake >= 1.16.4

ffontaine:
Drop call to `AM_PROG_AR` which will raise the following build failure with automake >= 1.16.4 because configure.ac doesn't call `AM_INIT_AUTOMAKE`:

`configure.ac: error: required file 'ar-lib' not found`

Adding a call to `AM_INIT_AUTOMAKE` will copy ar-lib file but will raise another build failure because tinydtls use hand-written `Makefile.in` instead of `Makefile.am`

Fixes:
 - http://autobuild.buildroot.org/results/c2fa1260cb4a2dbf2d7a90fea3ba99a389bcd470

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -29,7 +29,6 @@ test -e install-sh || touch install-sh
 AC_PROG_MAKE_SET
 AC_PROG_CC
 AC_PROG_RANLIB
-m4_ifdef([AM_PROG_AR], [AM_PROG_AR], [AR=ar])
 
 AC_C_BIGENDIAN
 


obgm:
Automake is not used by intention, so adding `AM_INIT_AUTOMAKE` would be wrong. Note that the check for `AM_PROG_AR` is there for a reason which is documented in the commit that has introduced this line. Dropping this line breaks the build because AR is not set. I am open for a better solution, but for now, commit 5c6fd17 fixes the issue with `ar-lib`. (Note that `autogen.sh` must be called to create ar-lib if it does not exist).

ffontaine:
OK, then I'm closing this PR. I'll update buildroot to touch `ar-lib`.

PR #104, opened 2021-10-10
Include CertificateRequest only for ECC builds.

boaks:
Add DTLS_ECC macros to include CertificateRequest processing only, if
build supports ECC.


boaks:
Hi @obgm ,

I think, this is an easy one, fast to merge :-).

PR #105, opened 2021-10-18
Add CCM cipher suite variants

boaks:
In case of CCM gets recommended and CCM8 deprecated.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -541,6 +541,42 @@ static uint8 compression_methods[] = {
   TLS_COMPRESSION_NULL
 };
 
+typedef enum {
+  DTLS_KEY_EXCHANGE_NONE,
+  DTLS_KEY_EXCHANGE_PSK,
+  DTLS_KEY_EXCHANGE_ECDHE_ECDSA
+} cipher_suite_key_exchange_algorithm_t;
+
+typedef struct cipher_suite_param_t {
+  uint16_t cipher_suite;
+  uint8_t mac_length;
+  cipher_suite_key_exchange_algorithm_t key_exchange_algorithm;
+} cipher_suite_param_t;
+
+static const struct cipher_suite_param_t cipher_suite_params[] = {
+#ifdef DTLS_PSK
+  { TLS_PSK_WITH_AES_128_CCM_8,          8, DTLS_KEY_EXCHANGE_PSK },
+  { TLS_PSK_WITH_AES_128_CCM,           16, DTLS_KEY_EXCHANGE_PSK },
+#endif /* DTLS_PSK */
+#ifdef DTLS_ECC
+  { TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8,  8, DTLS_KEY_EXCHANGE_ECDHE_ECDSA },
+  { TLS_ECDHE_ECDSA_WITH_AES_128_CCM,   16, DTLS_KEY_EXCHANGE_ECDHE_ECDSA },
+#endif /* DTLS_ECC */
+  /* TLS_NULL_WITH_NULL_NULL must always be the last entry as it
+   * indicates the stop marker for the traversal of this table. */
+  { TLS_NULL_WITH_NULL_NULL,             0, DTLS_KEY_EXCHANGE_NONE }
+ };
+
+static inline const cipher_suite_param_t* get_cipher_suite_param(dtls_cipher_t cipher)
+{
+  const cipher_suite_param_t *param = cipher_suite_params;
+  while ((param->cipher_suite != cipher) &&
+         (param->cipher_suite != TLS_NULL_WITH_NULL_NULL)) {
+    param++;
+  }
+  return param;
+}
+
 /**
  * Get MAC length of cipher suite.
  * \param cipher cipher suite
@@ -548,43 +584,19 @@ static uint8 compression_methods[] = {
  */
 static inline int get_cipher_suite_mac_len(dtls_cipher_t cipher)
 {
-#ifdef DTLS_ECC
-  if (cipher == TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8) {
-    return 8;
-  } else if (cipher == TLS_ECDHE_ECDSA_WITH_AES_128_CCM) {
-    return 16;
-  }
-#endif /* DTLS_ECC */
-#ifdef DTLS_PSK
-  if (cipher == TLS_PSK_WITH_AES_128_CCM_8) {
-    return 8;
-  } else if (cipher == TLS_PSK_WITH_AES_128_CCM) {
-    return 16;
-  }
-#endif /* DTLS_PSK */
-  return 0;
+  return get_cipher_suite_param(cipher)->mac_length;
 }
 
 /** returns true if the cipher matches TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8 or TLS_ECDHE_ECDSA_WITH_AES_128_CCM */
 static inline int is_tls_ecdhe_ecdsa_with_aes_128_ccm_x(dtls_cipher_t cipher)
 {
-#ifdef DTLS_ECC
-  return cipher == TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8 || cipher == TLS_ECDHE_ECDSA_WITH_AES_128_CCM;
-#else
-  (void) cipher;
-  return 0;
-#endif /* DTLS_ECC */
+  return get_cipher_suite_param(cipher)->key_exchange_algorithm == DTLS_KEY_EXCHANGE_ECDHE_ECDSA;
 }
 
 /** returns true if the cipher matches TLS_PSK_WITH_AES_128_CCM_8 or TLS_PSK_WITH_AES_128_CCM */
 static inline int is_tls_psk_with_aes_128_ccm_x(dtls_cipher_t cipher)
 {
-#ifdef DTLS_PSK
-  return cipher == TLS_PSK_WITH_AES_128_CCM_8 || cipher == TLS_PSK_WITH_AES_128_CCM;
-#else
-  (void) cipher;
-  return 0;
-#endif /* DTLS_PSK */
+  return get_cipher_suite_param(cipher)->key_exchange_algorithm == DTLS_KEY_EXCHANGE_PSK;
 }
 
 /** returns true if the application is configured for psk */
@@ -740,6 +752,7 @@ calculate_key_block(dtls_context_t *ctx,
   int pre_master_len = 0;
   dtls_security_parameters_t *security = dtls_security_params_next(peer);
   uint8 master_secret[DTLS_MASTER_SECRET_LENGTH];
+  cipher_suite_key_exchange_algorithm_t key_exchange_algorithm;
   (void)role; /* The macro dtls_kb_size() does not use role. */
 
   if (!security) {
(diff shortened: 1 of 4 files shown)


dtls.c
```diff
@@ -542,25 +542,48 @@ static uint8 compression_methods[] = {
...
-/** returns true if the cipher matches TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8 */
-static inline int is_tls_ecdhe_ecdsa_with_aes_128_ccm_8(dtls_cipher_t cipher)
+/**
+ * Get MAC length of cipher suite.
+ * \param cipher cipher suite
+ * \return MAC length of cipher. \c 0, if cipher is not supported.
+ */
+static inline int get_cipher_suite_mac_len(dtls_cipher_t cipher)
+{
+#ifdef DTLS_ECC
```
obgm, line 552: For a little more future flexibility, I suggest restructuring this code to set a local variable which is then returned. This could go into a switch statement, e.g.  (not tested):

```c
int length;
switch (cipher) {
default:
  length = 0;
  break;
#ifdef DTLS_ECC
case TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8:
  length = 8;
  break;
case TLS_ECDHE_ECDSA_WITH_AES_128_CCM:
  length = 16;
  break;
#endif /* DTLS_ECC */
#ifdef DTLS_PSK
case TLS_PSK_WITH_AES_128_CCM_8:
  length = 8;
  break;
case TLS_PSK_WITH_AES_128_CCM:
  length = 16;
  break;
#endif /* DTLS_PSK */
}
return length;
```

Or, even more flexible, with a table (not tested; `uint8_t` used to represent the size of the MAC tag to match the `tag_length` field of `dtls_ccm_params_t`):

```c
static inline uint8_t get_cipher_suite_mac_len(dtls_cipher_t cipher) {
  struct cipher_suite_param_t {
    uint16_t cipher_suite;
    uint8_t mac_length;
  };
  const struct cipher_suite_param_t cipher_suite_params[] = {
#ifdef DTLS_PSK
    { TLS_PSK_WITH_AES_128_CCM_8,          8 },
    { TLS_PSK_WITH_AES_128_CCM,           16 },
#endif /* DTLS_PSK */
#ifdef DTLS_ECC
    { TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8,  8 },
    { TLS_ECDHE_ECDSA_WITH_AES_128_CCM,   16 },
#endif /* DTLS_ECC */
    /* TLS_NULL_WITH_NULL_NULL must always be the last entry as it
     * indicates the stop marker for the traversal of this table. */
    { TLS_NULL_WITH_NULL_NULL,             0 }
  };

  const struct cipher_suite_param_t *p = cipher_suite_params;
  while ((p->cipher_suite != cipher) &&
         (p->cipher_suite != TLS_NULL_WITH_NULL_NULL)) {
    p++;
  }
  return p->mac_length;
}
```

dtls.c
```diff
@@ -1416,24 +1444,27 @@ dtls_prepare_record(dtls_peer_t *peer, dtls_security_parameters_t *security,
...
-  } else { /* TLS_PSK_WITH_AES_128_CCM_8 or TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8 */
+  } else { /* TLS_PSK_WITH_AES_128_CCM_8, TLS_PSK_WITH_AES_128_CCM, TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8 or TLS_ECDHE_ECDSA_WITH_AES_128_CCM */
     /**
      * length of additional_data for the AEAD cipher which consists of
      * seq_num(2+6) + type(1) + version(2) + length(2)
      */
 #define A_DATA_LEN 13
     unsigned char nonce[DTLS_CCM_BLOCKSIZE];
     unsigned char A_DATA[A_DATA_LEN];
+    int macLen = get_cipher_suite_mac_len(security->cipher);
```
obgm, line 1455: With previous suggested change option 2, this type now should probably be `uint8_t`.

And, as this is C code, please follow the usual conventions for variable naming, i.e., `mac_len`.

dtls.c
```diff
@@ -3358,17 +3397,18 @@ decrypt_verify(dtls_peer_t *peer, uint8 *packet, size_t length,
...
```
obgm, line 3408: See previous comment.

dtls.c
```diff
@@ -3358,17 +3397,18 @@ decrypt_verify(dtls_peer_t *peer, uint8 *packet, size_t length,
...
     /* For backwards-compatibility, dtls_encrypt_params is called with
      * M=<macLen> and L=3. */
-    const dtls_ccm_params_t params = { nonce, 8, 3 };
+    const dtls_ccm_params_t params = { nonce, macLen, 3 };
 
     if (clen < 16)		/* need at least IV and MAC */
```
obgm, line 3413: I think that now, we might need to have to check for something like `clen < (mac_len + MAX_IV_LENGTH)`, with `MAX_IV_LENGTH` currently being 8. 
Otherwise, the newly added `clen - macLen` below might yield an underflow.

dtls.c
boaks: Ok

boaks:
The first part of the changes. I will provided the rest at the weekend.

dtls.c
boaks: done

dtls.c
boaks: done

dtls.c
boaks: done.
(there is no symbolic IV-length, right?)

obgm: Correct—for some unknown reason, that has not been defined when crafting this check.

boaks: I would propose to go without for this PR.
And have a second PR introducing [RECORD_IV_LENGTH](https://datatracker.ietf.org/doc/html/rfc5246#page-25). 

WDYT?

obgm: Sure, a separate PR would be the cleanest  solution.

dtls.c
boaks: According the 1. proposal, the switch over enum with internal `#ifdef` results in warnings. Therefore I assumed, that for this easy case, the simple `if` are a not to bad solution.
For future extensions you right,  that a parameter table isn't that bad.

Should I amend this PR?

boaks:
I added a `cipher_suite_param_t` in a second commit. Is extended the original proposal by psk and ecdhe_ecdsa and used it for the other functions.

dtls.c
```diff
@@ -541,50 +541,62 @@ static uint8 compression_methods[] = {
   TLS_COMPRESSION_NULL
 };
 
+typedef enum {
+  NONE,
```
obgm, line 545: I wonder if we should prefix these with, e.g. `DTLS_CIPHER_` or similar. Especially `NONE` might be prone to name clashes.

boaks: What's about `KEY_EXCHANGE_...`?

obgm:
There is a lot of whitespace change which makes reading this PR a bit difficult.
Anyway, what I liked about the previous approach was that a lot of the conditionals were resolved at compile time. I do not know how good the optimizer is in unrolling the check loop.

dtls.c
obgm: WFM

boaks:
> There is a lot of whitespace change which makes reading this PR a bit difficult.

I'm used to fix the format of functions when I change them. Yes, that makes it hard to compare. One idea would then be, to spend some time in fix the format at once, and the rebase this PR. Though many files shows "Eclipse Public License v1.0" which is replaced by v2.0, such a cleanup seems to be required anyway.

Should I go for format and license fix next week?  

> what I liked about the previous approach was that a lot of the conditionals were resolved at compile time

```
static inline int is_tls_ecdhe_ecdsa_with_aes_128_ccm_8(dtls_cipher_t cipher)
{
#ifdef DTLS_ECC
  return cipher == TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8;
#else
  (void) cipher;
  return 0;
#endif /* DTLS_ECC */
}
```

The point seems to be, that using a reduced variety, makes it also possible to use very simple implementations.
Maybe, we add the "key_exchange_type to the handshake data and process it only once?

boaks:
> Maybe, we add the key_exchange_type to the handshake data and process it only once?

After analyzing it again, maybe exchange the `dtls_handshake_parameters_t.cipher` with `dtls_handshake_parameters_t.cipher_suite_params_index` as index into `cipher_suite_params` makes it simple again.

```
static inline int is_tls_ecdhe_ecdsa_with_aes_128_ccm_8(int cipher_suite_params_index)
{
  return DTLS_KEY_EXCHANGE_ECDHE_ECDSA == cipher_suite_params[cipher_suite_params_index].key_exchange_algorithm;
}
```

boaks:
I reverted many of the whitespace changes, I consider that is easier done separately.
I added also the "cipher_index" in order to obsolete the repetition of the loops.
I remove the is `is_tls_???_with_aes_128_ccm_???` and repaced the by comparing the key exchange algorithm.
PR #113 will adapted, once it gets clear, which variant is wanted.

dtls.c
```diff
@@ -562,41 +563,51 @@ static const struct cipher_suite_param_t cipher_suite_params[] = {
   { TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8,  8, DTLS_KEY_EXCHANGE_ECDHE_ECDSA },
   { TLS_ECDHE_ECDSA_WITH_AES_128_CCM,   16, DTLS_KEY_EXCHANGE_ECDHE_ECDSA },
 #endif /* DTLS_ECC */
-  /* TLS_NULL_WITH_NULL_NULL must always be the last entry as it
-   * indicates the stop marker for the traversal of this table. */
-  { TLS_NULL_WITH_NULL_NULL,             0, DTLS_KEY_EXCHANGE_NONE }
  };
 
-static inline const cipher_suite_param_t* get_cipher_suite_param(dtls_cipher_t cipher)
+static const int last_cipher_suite_param = sizeof(cipher_suite_params) / sizeof(cipher_suite_param_t);
```
obgm, line 568: The datatype for `last_cipher_suite_param` must be `size_t`.

dtls.c
```diff
@@ -562,41 +563,51 @@ static const struct cipher_suite_param_t cipher_suite_params[] = {
...
+
+static inline int get_cipher_index(dtls_cipher_t cipher)
```
obgm, line 570: The return type must be `size_t`.

dtls.c
```diff
@@ -562,41 +563,51 @@ static const struct cipher_suite_param_t cipher_suite_params[] = {
...
 /**
- * Get MAC length of cipher suite.
- * \param cipher cipher suite
- * \return MAC length of cipher. \c 0, if cipher is not supported.
+ * Get cipher suite.
+ * \param cipher_index index to cipher suite params
+ * \return cipher suite.
  */
-static inline int get_cipher_suite_mac_len(dtls_cipher_t cipher)
+static inline dtls_cipher_t get_cipher_suite(int cipher_index)
```
obgm, line 585: `cipher_index` must be of type `size_t`.

dtls.c
```diff
@@ -562,41 +563,51 @@ static const struct cipher_suite_param_t cipher_suite_params[] = {
...
 {
-  return get_cipher_suite_param(cipher)->mac_length;
+  assert(0 <= cipher_index && cipher_index < last_cipher_suite_param);
```
obgm, line 587: `0 <= cipher_index` can be omitted when the data types for the index have been changed to unsigned types.

dtls.c
```diff
@@ -562,41 +563,51 @@ static const struct cipher_suite_param_t cipher_suite_params[] = {
...
+  return cipher_suite_params[cipher_index].cipher_suite;
 }
 
-/** returns true if the cipher matches TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8 or TLS_ECDHE_ECDSA_WITH_AES_128_CCM */
-static inline int is_tls_ecdhe_ecdsa_with_aes_128_ccm_x(dtls_cipher_t cipher)
+/**
+ * Get key exchange algorithm of ciphher suite.
```
obgm, line 592: "ciphher" → "cipher"

obgm (commented):
I like the new approach with the index into the cipher_suite_params table. I have only a few comments, primarily to fix the data types.

dtls.c
```diff
@@ -814,12 +822,9 @@ calculate_key_block(dtls_context_t *ctx,
     return dtls_alert_fatal_create(DTLS_ALERT_INTERNAL_ERROR);
 #endif /* DTLS_ECC */
   case DTLS_KEY_EXCHANGE_NONE:
-    assert(!"calculate_key_block: tried to use NULL cipher\n");
-    return dtls_alert_fatal_create(DTLS_ALERT_INSUFFICIENT_SECURITY);
-
   default:
-    dtls_crit("calculate_key_block: unknown key exchange %d\n", key_exchange_algorithm);
-    return dtls_alert_fatal_create(DTLS_ALERT_INTERNAL_ERROR);
+    assert(!"calculate_key_block: not supported key exchange algorithm\n");
```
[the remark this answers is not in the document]
boaks: Let me try to explain my view:

The most critical "code points" here are the cipher-suite code-points in the Hellos. These code-points must be contained in the `cipher_suite_params`. The incoming/received "unknown" is therefore the cipher-suite and that fails ahead.
When this "internal/static" function "calculate_key_block" is called, only the  `cipher_suite_key_exchange_algorithm_t` contained in that table are possible.
The "switch-default" itself (previous "unknown key exchange") is to suppress compiler warnings, it doesn't occur (unknown cipher suites are detected ahead and unknown key exchanges don't exist on their own).
Therefore I just moved it to the `DTLS_KEY_EXCHANGE_NONE` in order to save a couple of bytes.

PR #106, opened 2021-11-18
Prepare for ccm

boaks:
Fix some logging statements.
Fix some misleading comments.
Use dtls_decrypt/encrypt with params to prepare for CCM additional to CCM8.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -790,7 +790,7 @@ calculate_key_block(dtls_context_t *ctx,
 #endif /* !DTLS_ECC */
 
   default:
-    dtls_crit("calculate_key_block: unknown cipher %x04 \n", handshake->cipher);
+    dtls_crit("calculate_key_block: unknown cipher %04x\n", handshake->cipher);
     return dtls_alert_fatal_create(DTLS_ALERT_INTERNAL_ERROR);
   }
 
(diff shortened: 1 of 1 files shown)


boaks:
This seems to be also a easy one.

dtls.c
```diff
@@ -1496,10 +1496,14 @@ dtls_prepare_record(dtls_peer_t *peer, dtls_security_parameters_t *security,
     memcpy(A_DATA + 8,  &DTLS_RECORD_HEADER(sendbuf)->content_type, 3); /* type and version */
     dtls_int_to_uint16(A_DATA + 11, res - 8); /* length */
 
-    res = dtls_encrypt(start + 8, res - 8, start + 8, nonce,
-		       dtls_kb_local_write_key(security, peer->role),
-		       dtls_kb_key_size(security, peer->role),
-		       A_DATA, A_DATA_LEN);
+    /* For backwards-compatibility, dtls_encrypt_params is called with
+     * M=<macLen> and L=3. */
+    const dtls_ccm_params_t params = { nonce, 8, 3 };
```
obgm, line 1501: To be backwards compatible to really old compilers, we have so far put variable definitions on the beginning of a block. `params` therefore should be moved to the beginning of the function body or enclosed together with the `dtls_encrypt_params()` invocation in `{` ... `}`.
The same holds below in `decrypt_verify()`.

obgm (commented):
I have a small comment, otherwise it is ready for merge.

dtls.c
boaks: OK, I adapt it.

boaks: The params are move. I removed the const as well, though the intention would be to place the macLength there.

boaks: if it's OK for you, I squash the commits.

boaks: I also adapted the "add_ccm" PR. It's possible to get the macLen ahead. so const will also work. I will change it again.

obgm: Ok, great. Please squash.

obgm:
When squashing, may I ask you to pick a more specific first line in the commit message that describes what this commit does in detail (prefixed by the name of the affected file: `dtls.c:`.

boaks:
sure ...

boaks:
> dtls.c: Minor fixes in logging and comments.

?

obgm:
Well, the major functional change here is the switch from `dtls_encrypt()` to `dtls_encrypt_params()`, IMO.

boaks:
OK, next try :-).

obgm:
Great, thank you very much!

PR #108, opened 2021-11-19
Client cid

boaks:
Start support DTLS Connection ID / RFC 9146.

Client implementation using an empty CID to indicate support.
If the server negotiates a CID, that is used to send records.
Setup matches the very common use-case, where the server has a fixed ip-address, but the client ip-address's may be temporary. Requires a server with cid support, which is not contained in this PR.

### DIFF
--- a/crypto.h
+++ b/crypto.h
@@ -102,6 +102,11 @@ typedef struct {
     uint64_t bitfield;
 } seqnum_t;
 
+/* Maximum CID length. */
+#ifndef DTLS_MAX_CID_LENGTH
+#define DTLS_MAX_CID_LENGTH 16
+#endif
+
 typedef struct {
   dtls_compression_t compression;	/**< compression method */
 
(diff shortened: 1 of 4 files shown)


boaks:
@obgm 

Would it be possible to include that into a special development branch, e.g. "develop_cid"? 
It would make it little easier for some to use it. Sure, it's also not that complicated either to fetch the PR or to fetch from the bosch-io fork. So it will also be OK, to leave it "as it is".

obgm:
Sure. There now is a new branch [feature/connection_id](/eclipse/tinydtls/tree/feature/connection_id) for this. (I picked the name to be rather explicit, whereas it is clear from the nature of this branch that it is for development.)

boaks:
@obgm 

Please "reset" the feature/connection_id branch to this Pr in order to include the recent bugfixes.

obgm:
Done, the branch head now points to [8ff2595](https://github.com/eclipse/tinydtls/pull/108/commits/8ff2595210f15e2b4aa946c3c27732a88fde2b24).

boaks:
I prepared PR #149  to merge it into main.

Issue #78, opened 2021-05-21
Client proceeds in handshake after receiving consecutive CertificateRequest messages


pfg666:
## Non-conformance Bug
**OS**  
Linux, Ubuntu 20
develop branch, commit c20c6323

**Expected behavior**
The client should not proceed with a handshake where it receives distinct consecutive CertificateRequest messages.

**Actual Behavior**
Our testing shows that the client upon receiving these messages proceeds with the handshake as normal; upon receiving ServerHelloDone, it generates the next flight of messages as can be seen in the capture below taken on my machine.

![etinydtls_multiple_certreq](https://user-images.githubusercontent.com/2325013/119117008-198c9080-ba29-11eb-8bd9-b057f4c84bc4.png)

Note that the later CertificateRequest is not a retransmission of the former. Its message_sequence value differs, so it should have been processed as a distinct handshake message.

**Steps to Reproduce**
I attached files necessary for reproduction using [DTLS-Fuzzer](https://github.com/assist-project/dtls-fuzzer/), a Java-based tool for testing DTLS libraries. Also included in the archive is a capture of the interaction show above. DTLS-Fuzzer requires  the JDK for Java 8. On Ubuntu, this can be installed  by running:
`sudo apt-get install openjdk-8-jdk`

Unpack the archive, `cd` to resulting folder and run `bash reproduce.sh`, while running an instance of Wireshark on the side. The reproduction script will: 
* setup eclipse TinyDTLS and DTLS-Fuzzer
* launch the TinyDTLS client using the 'tests/dtls-client utility;
* launch DTLS-Fuzzer to execute the input sequence 'test_sequence'  which exposes the bug.

Thanks!

[reproduction.tar.gz](https://github.com/eclipse/tinydtls/files/6521626/reproduction.tar.gz)

mrdeep1:
@pfg666 With your script, I get the following failure (I had to do `sudo apt install maven` and then build everything from scratch)
````
13:38:18 [main] INFO : Main - Noticed that the first argument is a file. Processing it as an argument file.
13:38:18 [main] INFO : Main - Processing command state-fuzzer-client
13:38:18 [main] INFO : Main - Debug operation is engaged
13:38:18 [main] INFO : Main - Running test runner
13:38:23 [main] WARN : WorkflowConfigurationFactory - Unsupported key exchange algorithm: NULL, not creating ClientKeyExchange Message
[...]

PR #109, opened 2021-11-22
dtls.c: Check for optional handshake messages

boaks:
In general the handshake state machine forces the handshakes to go
through
a defined order. But optional handshake message are not controlled by
that.

This fix introduces a optional handshake message field. With that, the
state-machine is enabled to check, if this message is (optionally)
expected. When processed, the field is reset and receiving the optional
message twice will therefore fail the handshake.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -3628,7 +3628,8 @@ handle_handshake_msg(dtls_context_t *ctx, dtls_peer_t *peer, uint8 *data, size_t
 #ifdef DTLS_ECC
   case DTLS_HT_CERTIFICATE_REQUEST:
 
-    if (state != DTLS_STATE_WAIT_SERVERHELLODONE) {
+    if (state != DTLS_STATE_WAIT_SERVERHELLODONE ||
+        !is_tls_ecdhe_ecdsa_with_aes_128_ccm_8(peer->handshake_params->cipher)) {
       return dtls_alert_fatal_create(DTLS_ALERT_UNEXPECTED_MESSAGE);
     }
 
--- a/dtls.c
+++ b/dtls.c
@@ -3440,6 +3440,7 @@ dtls_renegotiate(dtls_context_t *ctx, const session_t *dst)
 
   peer->handshake_params->hs_state.mseq_r = 0;
   peer->handshake_params->hs_state.mseq_s = 0;
+  peer->optional_handshake_message = DTLS_HT_NO_OPTIONAL_MESSAGE;
 
   if (peer->role == DTLS_CLIENT) {
     /* send ClientHello with empty Cookie */
@@ -3553,8 +3554,10 @@ handle_handshake_msg(dtls_context_t *ctx, dtls_peer_t *peer, uint8 *data, size_t
     }
     if (is_tls_ecdhe_ecdsa_with_aes_128_ccm_8(peer->handshake_params->cipher))
       peer->state = DTLS_STATE_WAIT_SERVERCERTIFICATE;
-    else
+    else {
+      peer->optional_handshake_message = DTLS_HT_SERVER_KEY_EXCHANGE;
       peer->state = DTLS_STATE_WAIT_SERVERHELLODONE;
+    }
     /* update_hs_hash(peer, data, data_length); */
 
     break;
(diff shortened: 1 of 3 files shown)


boaks:
Alternative proposal for #88

boaks:
This one would be my favorite for the next merge :-).
It fixes #78 and obsoletes #88.

boaks:
@obgm

That 20 lines would close three topics, #78, #88 and this PR itself :-).

peer.h
```diff
@@ -52,6 +57,7 @@ typedef struct dtls_peer_t {
 
   dtls_peer_type role;       /**< denotes if this host is DTLS_CLIENT or DTLS_SERVER */
   dtls_state_t state;        /**< DTLS engine state */
+  int16_t optionalHandshakeMessage; /**< optional next handshake message, DTLS_HT_NO_OPTIONAL_MESSAGE, if no optional message is expected. */
```
obgm, line 60: Please follow C coding conventions. An identifier thus would merely look like this: `optional_handshake_message`.

dtls.c
```diff
@@ -3628,10 +3633,11 @@ handle_handshake_msg(dtls_context_t *ctx, dtls_peer_t *peer, uint8 *data, size_t
 #ifdef DTLS_ECC
   case DTLS_HT_CERTIFICATE_REQUEST:
 
-    if (state != DTLS_STATE_WAIT_SERVERHELLODONE) {
+    if (!is_tls_ecdhe_ecdsa_with_aes_128_ccm_8(peer->handshake_params->cipher) ||
```
obgm, line 3636: Hm, the newly introduced first check seems to be unrelated to this PR. This might warrant at least a separate commit.

obgm (changes requested):
Thank you for this fix. I wonder why `DTLS_HT_NO_OPTIONAL_MESSAGE` is defined in *peer.h* rather than in *dtls.c* which is the only place where it is used. (The other `DTLS_HT_` types are defined in _dtls.h_.)
Moreover, as the PR description indicates, this flag directly relates to the handshake state. Therefore, I think it should go into the `dtls_handshake_parameters_t` structure (possibly even as an enum?).

I also suggest some minor changes to the commit message: First, lines longer than 72 for characters are rather unusual and lead to the stray line breaks as seen in the GitHub presentation. Second, I suggest slight rewording to avoid too much passive tense, e.g., as shown below (with punctuation error fixed):

```
dtls.c: Check for optional handshake messages

In general the handshake state machine forces the handshakes to go
through a defined order. But optional handshake message are not
controlled by that.

This fix introduces a optional handshake message field. This new
field enables the state-machine to check if this message
is (optionally) expected. When processed, the field is reset and
receiving the optional message twice will therefore fail the
handshake.


Fixes: Issue #78

```

peer.h
boaks: Too much java ...

dtls.c
boaks: If you prefer it, sure.

boaks:
> I wonder why DTLS_HT_NO_OPTIONAL_MESSAGE is defined in peer.h rather than in dtls.c which is the only place where it is used. (The other DTLS_HT_ types are defined in dtls.h.)

My intention was to define it close to the field, where the value is used. I can move it to dtls.h.

boaks:
Next try. 
(If the two empty lines before "Fixes: Issue " intended? I reduced it to one, but if it's intended, I can add a additional one.)

obgm:
> Next try. (Is the two empty lines before "Fixes: Issue " intended? I reduced it to one, but if it's intended, I can add a additional one.)

Sorry, this was a copy&paste bug. Thanks for fixing this right away.

obgm:
Thanks for the changes which look good so far.
The only open question now is whether the new flag `optional_handshake_message` should be in the `dtls_handshake_parameters_t` structure?

boaks:
> The only open question now is whether the new flag optional_handshake_message should be in the dtls_handshake_parameters_t structure?

My intention to put it in `dtls_peer_t` was to keep it close to `dtls_state_t state`, though for me both are related.
Anyway, if you prefer, it could be moved to `dtls_handshake_parameters_t`.

obgm:
Agreed: Keeping it close to state may make even more sense than close to hs_state. My bad.

obgm (approved):

boaks:
I considered the other variant as well. It saves 2 bytes per peer. In my opinion, it's not that easy to understand, why that is required and working, so I gone for have it close to the related state.

Thanks for merging.

PR #111, opened 2021-12-10
dtls.c: Remove unused security variable

mrdeep1:


boaks:
Nice catch ;-).

PR #116, opened 2021-12-22
dtls.c: use dtls_security_params_read_epoch when decode record.

boaks:
Adapt the test of the return value of decrypt_verify
be larger than 0 on all places.


boaks:
@obgm 

If you have time. These two fixes are very simple. Should be possible to get fast reviewed.

PR #118, opened 2021-12-24
dtls.c: fix memory leak in handshake message queue.

boaks:
The received message is cloned in order to store it in the handshake
message queue. If the message is then processed delayed, removing it is
not enough. The node must be freed in order to free the cloned message
as well.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -4015,6 +4015,10 @@ handle_handshake(dtls_context_t *ctx, dtls_peer_t *peer, uint8 *data, size_t dat
           netq_remove(&peer->handshake_params->reorder_queue, node);
           next = 1;
           res = handle_handshake_msg(ctx, peer, node->data, node->length);
+
+          /* free message data */
+          netq_node_free(node);
+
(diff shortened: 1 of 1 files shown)


dtls.c
```diff
@@ -4015,6 +4015,10 @@ handle_handshake(dtls_context_t *ctx, dtls_peer_t *peer, uint8 *data, size_t dat
           netq_remove(&peer->handshake_params->reorder_queue, node);
           next = 1;
           res = handle_handshake_msg(ctx, peer, node->data, node->length);
+
+          // free message data
```
obgm, line 4019: I know, it's pendantic: But could you please use the `/* ... */` comment form even for single-line comments?

obgm (commented):
Thanks, looks good, except for one minor cosmetic change.

dtls.c
boaks: Sure :-)

boaks: Done.

PR #121, opened 2022-01-07
dtls_debug.c: fix minor bug with zero termination.

boaks:
Providing a too small buffer may result in missing the zero termination.
This fix prevents that, even if currently no too small buffers are
used.

### DIFF
--- a/dtls_debug.c
+++ b/dtls_debug.c
@@ -107,34 +107,44 @@ dtls_strnlen(const char *s, size_t maxlen) {
   return n;
 }
 
+/**
+ * Write service-address as text to buffer.
+ *
+ * The text is \000 terminated.
+ *
+ * \param addr   The session including the sockaddr.
+ * \param buf    The buffer to write the service address into.
+ * \param len    The actual length of \p buf.
+ * \return Less than zero on error, the number of bytes written otherwise.
+ */
 static size_t
 dsrv_print_addr(const session_t *addr, char *buf, size_t len) {
 #ifdef HAVE_ARPA_INET_H
   const void *addrptr = NULL;
   in_port_t port;
   char *p = buf;
+  size_t append;
+  int err;
+
+  /* max. '[' ipv6 ']:' port '\0' */
+  assert(len >= 1 + INET6_ADDRSTRLEN + 2 + 5 + 1);
 
   switch (addr->addr.sa.sa_family) {
   case AF_INET:
-    if (len < INET_ADDRSTRLEN)
-      return 0;
-
     addrptr = &addr->addr.sin.sin_addr;
     port = ntohs(addr->addr.sin.sin_port);
     break;
   case AF_INET6:
-    if (len < INET6_ADDRSTRLEN + 2)
-      return 0;
-
     *p++ = '[';
-
+    --len;
     addrptr = &addr->addr.sin6.sin6_addr;
     port = ntohs(addr->addr.sin6.sin6_port);
-
     break;
   default:
-    memcpy(buf, "(unknown address type)", min(22, len));
-    return min(22, len);
+    /* include terminating \000 */
+    append = strlen("(unknown address type)");
+    memcpy(p, "(unknown address type)", append + 1);
+    return append;
   }
 
   if (inet_ntop(addr->addr.sa.sa_family, addrptr, p, len) == 0) {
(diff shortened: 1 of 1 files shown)


dtls_debug.c
```diff
@@ -107,12 +107,23 @@ dtls_strnlen(const char *s, size_t maxlen) {
...
+ * \param len    The actual length of \p buf.
+ * \return Less than zero on error, the number of bytes written otherwise.
+ */
 static size_t
 dsrv_print_addr(const session_t *addr, char *buf, size_t len) {
 #ifdef HAVE_ARPA_INET_H
   const void *addrptr = NULL;
   in_port_t port;
   char *p = buf;
+  int append;
```
obgm, line 126: This should be of type `size_t`.

dtls_debug.c
```diff
@@ -127,31 +138,43 @@ dsrv_print_addr(const session_t *addr, char *buf, size_t len) {
...
+    --len;
 
     addrptr = &addr->addr.sin6.sin6_addr;
     port = ntohs(addr->addr.sin6.sin6_port);
 
     break;
   default:
-    memcpy(buf, "(unknown address type)", min(22, len));
-    return min(22, len);
+    // include terminating \000
```
obgm, line 148: Please use the `/* ... */` comment syntax.

dtls_debug.c
```diff
@@ -127,31 +138,43 @@ dsrv_print_addr(const session_t *addr, char *buf, size_t len) {
...
+    return append;
   }
 
   if (inet_ntop(addr->addr.sa.sa_family, addrptr, p, len) == 0) {
     perror("dsrv_print_addr");
     return 0;
   }
 
-  p += dtls_strnlen(p, len);
+  // append inet_ntop to p
```
obgm, line 162: Please use the `/* ... */` comment syntax.

dtls_debug.c
```diff
@@ -127,31 +138,43 @@ dsrv_print_addr(const session_t *addr, char *buf, size_t len) {
...
     port = ntohs(addr->addr.sin6.sin6_port);
 
     break;
   default:
-    memcpy(buf, "(unknown address type)", min(22, len));
-    return min(22, len);
+    // include terminating \000
+    append = min(23, len);
+    memcpy(p, "(unknown address type)", append);
+    if (len < 23) {
```
obgm, line 151: I wonder why the default case needs zero-termination while the other cases may not be zero-terminated (and do not need to, hence the returned length).

dtls_debug.c
boaks: Yes, I will do. (Sorry for that, I work on too much stuff in parallel with all different rules. Please keep on requesting to comply to your source format rules.)

dtls_debug.c
boaks: I consider the code to return a filled buffer only here, or after the snprintf.
And, maybe I'm wrong, I considered the snprintf to always terminate with 0.

dtls_debug.c
boaks: Done

dtls_debug.c
boaks: Done

dtls_debug.c
boaks: `snprintf`, returns `int`, therefore I chose that.

dtls_debug.c
mrdeep1: snprintf() does always include an explicit 0.  However, you cannot rely on the returned value as this could be oversize - it needs to be compared with the actual (length -1) provided to snprint() to prevent an overwrite of the provided buffer.

In the past, I have used strlen() to determine where the actual output has reached.

boaks: Without overflow, it adds the 0. But with overflow, it's required to decide, if the output should fail, or the port number gets truncated. Anyway, I guess, I must check on all steps, if there is an overflow ...

mrdeep1: With overflow, it still adds the 0 as the last byte in the available buffer as indicated in (as well as in my experience of)

> The functions snprintf() and vsnprintf() do not write more than size bytes 
> (including the terminating null byte ('\0')).

mrdeep1: Maybe add in an assert() if there is insuffient space so that the buffer size issue can easily be caught (rather than just through code inspection) and the size adjusted accordingly.

boaks: > With overflow, it still adds the 0 as the last byte

Hm, that was also, what I remembered. I try to check the docu, but maybe I still misunderstood it :-).

boaks: > Maybe add in an assert() 

Not that bad. Then maybe, just at the start checking for the min size of the buffer?

mrdeep1: Works for me

dtls_debug.c
boaks: I'm still in doubt, if `size_t` is really better. Or if two locals are required `int` and `size_t`.
Currently I don't see, that the `int` range is exceeded. I also don't see, that `snprintf` returns an error.
My preference is still `int`. but if the other variants, `size_t` or `int` and `size_t` are preferred, I adapt the code.

obgm: Well, for one, `len` has the type `size_t`, clearly documenting that `len` is for sizes. `append` pretty much deals with sizes as well and relates to `len` in this function. `dtls_strnlen()` returns a `size_t`, hence the result must be treated as `size_t`.
The `snprintf` return type indeed is of type `int`. As we do not know if and why some exotic implementations might to actually return an error here, we should use a second variable of type `int` for this.

boaks: OK, then go for two locals.

boaks:
Now, with two locals.

obgm:
Thanks for doing the changes!

PR #122, opened 2022-01-07
dtls_time.h: conditionally include time header files.

boaks:
Include time.h and sys/time.h based on HAVE_TIME_H and HAVE_SYS_TIME_H.

A first port for zephyr shows, that accidentally included time headers cause some other headers (socket) to fail. Easiest work-around is to switch these includes using the available macros. 

### DIFF
--- a/dtls_time.h
+++ b/dtls_time.h
@@ -23,7 +23,10 @@
 #define _DTLS_DTLS_TIME_H_
 
 #include <stdint.h>
+
+#ifdef HAVE_SYS_TIME_H
 #include <sys/time.h>
+#endif /* HAVE_SYS_TIME_H */
 
 #include "tinydtls.h"
 
@@ -50,7 +53,9 @@
 typedef uint64_t clock_time_t;
 #else /* WITH_CONTIKI || RIOT_VERSION */
 
+#ifdef HAVE_TIME_H
 #include <time.h>
+#endif /* HAVE_TIME_H */
 
 #ifndef CLOCK_SECOND
 # define CLOCK_SECOND 1000
--- a/sha2/sha2speed.c
+++ b/sha2/sha2speed.c
@@ -35,7 +35,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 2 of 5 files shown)


boaks:
Maybe a advanced issue:
tinyDtls uses time for to things, retransmission and timeouts, and logging.
For the the retransmission, a "monotone-uptime" seems to be relevant.
For logging more the calendar time, if available.

I'm currently not sure, what the "preferred portability pattern" is.
I see:
- [dtls_prng](https://github.com/eclipse/tinydtls/blob/develop/dtls_prng.c) and [platform-specific](https://github.com/eclipse/tinydtls/tree/develop/platform-specific) for the random.
- [dtls_time.c](https://github.com/eclipse/tinydtls/blob/develop/dtls_time.c) for uptime
- [dtls_debug.c](https://github.com/eclipse/tinydtls/blob/develop/dtls_debug.c) for logging

so which pattern should be considered?
Or a different idea?

I ask that, because I have already potted it to zephyr. Very easy, but I'm not sure, if an additional port makes currently really sense.

obgm:
Regarding this specific PR, I think that the test programs do not have the conditionals because they are supposed to work on POSIX only. (There also is no harm in having them, though.)

boaks:
> the test programs do not have the conditionals because they are supposed to work on POSIX only.

For me dtls_client and dtls_server are also good starting points to test a port.
At least I used the dtls_client for zephyr.

boaks:
@obgm

I would welcome, to have it merged. Is that possible for you?

obgm:
Sure!  I will address the related question as a separate issue.

boaks:
Thanks, I would work on a proposal for "portability" based on your comments above.
But currently my feeling is, the pending PRs should be processed first.
(I agree, that the limitation in time should not turn in a limitation in quality.)

My preference list (fix bugs first):
- PR #121
- PR #117
- PR #112
- PR #96

If that is done, I would "rebase" the CID and CCM branch and spend more time into the CCM to split the changes more on separate commits. But for now, I guess the list above is work enough :-).

PR #124, opened 2022-02-03
Split dtls_time into platform-specifc (WIP)

obgm:
This is a tracking issue for the separation of platform-specific functions for time measurement as raised in PR #122


boaks:
Maybe you can check #123 ?

obgm:
Yes, PR #123 it is on the list.

boaks:
PR #123 will somehow overlap this one.
And, if the type gets truncated to 32 bits, I think my remark about the overflow needs then to be considered.

obgm:
Likely. But I suggest discussing it in the affected PR.

PR #126, opened 2022-02-27
Add Windows support

JKRhb:
In order to make tinydtls more cross-platform compatible, this PR adds Windows support by conditionally including header files, an additional PRNG implementation, and missing definitions. This way tinydtls can be cross-compiled for Windows using MinGW, for example.

Some workarounds (type definitions in `sha2.h` and alternative included headers in `tests/dtls-server.c`) have been added in order to be able to actually compile the library. Using MinGW, some errors/warnings are thrown because a number of definitions from a libary called `Ws2_32.lib` cannot be found, but the resulting `libtinydtls.so` can be used as a `.dll` file under Windows nonetheless.

By running, for example, `./configure --host x86_64-w64-mingw32` you can generate Makefiles to test the cross-compilation on a Linux machine. MinGW needs to be installed to perform the actual compilation.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -213,7 +213,7 @@ static inline void free_context(dtls_context_t *context) {
 }
 #endif /* RIOT_VERSION */
 
-#ifdef WITH_POSIX
+#if defined(WITH_POSIX) || defined(IS_WINDOWS)
 
 static inline dtls_context_t *
 malloc_context(void) {
@@ -1610,8 +1610,10 @@ dtls_0_send_hello_verify_request(dtls_context_t *ctx,
 			     dtls_ephemeral_peer_t *ephemeral_peer,
 			     uint8 *data, size_t data_length)
 {
-  uint8 buf[DTLS_RH_LENGTH + DTLS_HS_LENGTH + data_length];
+  uint8 buf[DTLS_RH_LENGTH + DTLS_HS_LENGTH + DTLS_HV_LENGTH + DTLS_COOKIE_LENGTH];
   uint8 *p = dtls_set_record_header(DTLS_CT_HANDSHAKE, 0, &(ephemeral_peer->rseq), buf);
+
+  assert(data_length == DTLS_HV_LENGTH + DTLS_COOKIE_LENGTH);
 
   /* Signal DTLS version 1.0 in the record layer of ClientHello and
    * HelloVerifyRequest handshake messages according to Section 4.2.1
--- a/dtls.h
+++ b/dtls.h
@@ -341,8 +341,16 @@ void dtls_check_retransmit(dtls_context_t *context, clock_time_t *next);
 #define DTLS_CT_HANDSHAKE          22
 #define DTLS_CT_APPLICATION_DATA   23
 
+#ifdef __GNUC__
+#define PACK( __Declaration__ ) __Declaration__ __attribute__((__packed__))
+#elif defined(_MSC_VER)
+#define PACK( __Declaration__ ) __pragma( pack(push, 1) ) __Declaration__ __pragma( pack(pop))
+#else
+#error "Structure packing is not available for the used compiler."
+#endif
+
 /** Generic header structure of the DTLS record layer. */
-typedef struct __attribute__((__packed__)) {
+PACK(typedef struct) {
   uint8 content_type;		/**< content type of the included message */
   uint16 version;		/**< Protocol version */
   uint16 epoch;		        /**< counter for cipher state changes */
@@ -371,7 +379,7 @@ typedef struct __attribute__((__packed__)) {
 #define DTLS_HT_NO_OPTIONAL_MESSAGE        -1
 
 /** Header structure for the DTLS handshake protocol. */
-typedef struct __attribute__((__packed__)) {
+PACK(typedef struct) {
   uint8 msg_type; /**< Type of handshake message  (one of DTLS_HT_) */
   uint24 length;  /**< length of this message */
   uint16 message_seq; 	/**< Message sequence number */
@@ -381,7 +389,7 @@ typedef struct __attribute__((__packed__)) {
 } dtls_handshake_header_t;
 
 /** Structure of the Client Hello message. */
-typedef struct __attribute__((__packed__)) {
+PACK(typedef struct) {
   uint16 version;	  /**< Client version */
   uint32 gmt_random;	  /**< GMT time of the random byte creation */
   unsigned char random[28];	/**< Client random bytes */
@@ -392,7 +400,7 @@ typedef struct __attribute__((__packed__)) {
 } dtls_client_hello_t;
 
 /** Structure of the Hello Verify Request. */
-typedef struct __attribute__((__packed__)) {
+PACK(typedef struct) {
   uint16 version;		/**< Server version */
   uint8 cookie_length;	/**< Length of the included cookie */
   uint8 cookie[];		/**< up to 32 bytes making up the cookie */
--- a/dtls_debug.c
+++ b/dtls_debug.c
@@ -222,12 +222,13 @@ dsrv_print_addr(const session_t *addr, char *buf, size_t len) {
 
 #endif /* WITH_CONTIKI */
 
-#ifdef RIOT_VERSION
+#if defined(RIOT_VERSION) || defined(IS_WINDOWS)
   /* FIXME: Switch to RIOT own DEBUG lines */
+  /* TODO: Check if inet_ntop can be used on Windows */
   (void) addr;
   (void) buf;
   (void) len;
-#endif /* RIOT_VERSION */
+#endif /* RIOT_VERSION || IS_WINDOWS */
 
 #ifdef WITH_POSIX
   /* TODO: output addresses manually */
@@ -240,7 +241,7 @@ dsrv_print_addr(const session_t *addr, char *buf, size_t len) {
 
 #endif /* NDEBUG */
 
-#ifndef WITH_CONTIKI
+#if !defined(WITH_CONTIKI) && !defined(_MSC_VER)
 void
 dsrv_log(log_t level, const char *format, ...) {
   static char timebuf[32];
--- a/dtls_debug.h
+++ b/dtls_debug.h
@@ -78,11 +78,11 @@ void dtls_set_log_level(log_t level);
  * level is below or equal to the log level that set by
  * set_log_level(). */
 #ifdef HAVE_VPRINTF
-#if (defined(__GNUC__))
+#if (defined(__GNUC__) && !defined(__MINGW32__))
 void dsrv_log(log_t level, const char *format, ...) __attribute__ ((format(printf, 2, 3)));
-#else /* !__GNUC__ */
+#else /* !__GNUC__ && !__MINGW32__ */
 void dsrv_log(log_t level, const char *format, ...);
-#endif /* !__GNUC__ */
+#endif /* !__GNUC__ && !__MINGW32__ */
 #else
 #define dsrv_log(level, format, ...) PRINTF(format, ##__VA_ARGS__)
 #endif
--- a/dtls_time.c
+++ b/dtls_time.c
@@ -22,6 +22,11 @@
 #include "tinydtls.h"
 #include "dtls_time.h"
 
+#ifdef _MSC_VER
+#include <time.h>
+#include <windows.h>
+#endif
+
 #ifdef WITH_CONTIKI
 clock_time_t dtls_clock_offset;
 
@@ -64,7 +69,7 @@ dtls_ticks(dtls_tick_t *t) {
   *t = k_uptime_get();
 }
 
-#elif defined(WITH_POSIX)
+#elif defined(WITH_POSIX) || defined(IS_WINDOWS)
 
 time_t dtls_clock_offset;
 
@@ -88,6 +93,11 @@ void dtls_ticks(dtls_tick_t *t) {
   gettimeofday(&tv, NULL);
   *t = (tv.tv_sec - dtls_clock_offset) * DTLS_TICKS_PER_SECOND 
     + (tv.tv_usec * DTLS_TICKS_PER_SECOND / 1000000);
+#elif defined(_MSC_VER)
+  SYSTEMTIME current_time;
+  GetSystemTime(&current_time);
+  *t = (current_time.wSecond - dtls_clock_offset) * DTLS_TICKS_PER_SECOND 
+    + (current_time.wMilliseconds * DTLS_TICKS_PER_SECOND / 1000);
 #else
 #error "clock not implemented"
 #endif
--- a/ecc/ecc.c
+++ b/ecc/ecc.c
@@ -39,6 +39,7 @@
 //big number functions
 #include "ecc.h"
 #include <string.h>
+#include <assert.h>
 
 static uint32_t add( const uint32_t *x, const uint32_t *y, uint32_t *result, uint8_t length){
 	uint64_t d = 0; //carry
@@ -157,7 +158,8 @@ static int fieldSub(const uint32_t *x, const uint32_t *y, const uint32_t *modulu
 //finite Field multiplication
 //32bit * 32bit = 64bit
 static int fieldMult(const uint32_t *x, const uint32_t *y, uint32_t *result, uint8_t length){
-	uint32_t temp[length * 2];
+	uint32_t temp[(arrayLength + 1) * 2];
+	assert(length <= arrayLength + 1);
 	setZero(temp, length * 2);
 	setZero(result, length * 2);
 	uint8_t k, n;
--- a/platform-specific/dtls_prng_win.c
+++ b/platform-specific/dtls_prng_win.c
@@ -1,0 +1,50 @@
+/*******************************************************************************
+ *
+ * Copyright (c) 2022 Jan Romann and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompanies this distribution.
+ *
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ *******************************************************************************/
+
+#include "tinydtls.h"
+#include "dtls_prng.h"
+#include "dtls_debug.h"
+
+#define _CRT_RAND_S
+
+#include <stdlib.h>
+
+#if defined(__MINGW32__)
+/** Missing function declaration for rand_s under MingW. */
+__declspec(dllimport) int __cdecl rand_s(unsigned int*);
+#endif
+
+/**
+ * Fills @p buf with @p len random bytes. Returns a non-zero
+ * value on error.
+ */
+int
+dtls_prng(unsigned char *buf, size_t len) {
+  errno_t err;
+  unsigned int number;
+  size_t klen = len;
+  while (len--) {
+    err = rand_s(&number);
+    if (err != 0) {
+      dtls_emerg("PRNG failed\n");
+      return err;
+    }
+    *buf++ = number & 0xFF;
+  }
+  return klen;
+}
+
+void
+dtls_prng_init(unsigned seed) {
+  srand(seed);
+}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,3 +62,15 @@ jobs:
       run: |
         cmake -Dmake_tests=ON .
         make
+
+  build-windows:
+    name: Build for Windows using CMake
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Build tinydtls
+      run: |
+        cmake -G "Unix Makefiles" .
+        make
+
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,7 @@ dtls_config.h
 *.cooja
 *.econotag
 *.eval-adf7xxxmb4z
+*.exe
 *.exp5438
 *.iris
 *.mbxxx
@@ -70,3 +71,23 @@ dtls_config.h
 CMakeCache.txt
 cmake_install.cmake
 CMakeFiles
+
+# Windows build files
+*.dll
+
+# Visual Studio Build Files
+*.VC.db*
+.vs
+*.obj
+*.tlog
+*.log
+*.recipe
+*.vcxproj
+*.sln
+*.pdb
+*.vcxproj.user
+*.vcxproj.filters
+*.vcxproj.FileListAbsolute.txt
+tinydtls.dir
+Debug
+Release
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,7 @@ option(make_tests "Make test programs and examples" OFF)
 if(NOT PLATFORM)
    # PLATFORM seems to be not used
    set(PLATFORM "posix" CACHE STRING "Choose platform." FORCE)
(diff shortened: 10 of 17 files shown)


obgm:
Thanks @JKRhb for this contribution! 
Where does the pre-processor symbol `IS_WINDOWS` get set? The reason I am asking is that some of these adaptations seem not to be specific to mingw but Windows in general. In some environments, we already have the pre-processor symbol `_WIN32` that could be used as an indicator for platform Windows.

(I will provide a few more review comments shortly.)

JKRhb:
Thank you already for the feedback, @obgm :)

> Where does the pre-processor symbol `IS_WINDOWS` get set? The reason I am asking is that some of these adaptations seem not to be specific to mingw but Windows in general. In some environments, we already have the pre-processor symbol `_WIN32` that could be used as an indicator for platform Windows.

`IS_WINDOWS` is currently being defined in `tinydtls.h` (see [here](https://github.com/eclipse/tinydtls/pull/126/files#diff-22c994b978965a8225b7c0fd5df7b993695f3802b32e07b7ba0c9d0f328a5f67R35)) like so:

```c
#if defined(_WIN32) || defined(_WIN64)
#define IS_WINDOWS 1
#endif
```

I defined this new symbol as I was a bit unsure if `_WIN32` is set in every environment[^1]. If someone should report that compilation fails in their environment, then the list of symbols being checked could easily be expanded. If `_WIN32` should be a "catch all" definition, then I guess we could replace `IS_WINDOWS` and simplify things a bit :)

[^1]: There is also `_WIN32_`, `__WIN32__`, `WIN32`, `__MINGW32__`, `__MINGW64__` etc.

obgm:
Okay. I like the aprpoach for `IS_WINDOWS` (just missed the part where it is set).

boaks:
If possible, could you check, if the updated cmake support (PR #132 ) works for you?

JKRhb:
> If possible, could you check, if the updated cmake support (PR #132 ) works for you?

Sure, will do :) Thank you for following up on the CMake PR!

JKRhb:
Thanks for the ping, @obgm! I just pushed my local state, where I made some adjustments that take the newly added CMake support into account and should also allow for compilation with Visual Studio (besides MingW/Cygwin). This, however, does require additional Windows-specific workarounds, some of which might be able to be factored out into their own files.

What is currently still missing are the Windows-specific macro definitions for mutexes, which I have to do a bit more research for. I also need to do some more testing with Visual Studio, which I am not really familiar with. I think the PR should be finished in the next couple of days, though, and could already be reviewed in its current state :)

obgm:
Good. @boaks The email address used for the references commit is not covered by an ECA.

boaks:
that commit with my previous Email is already merged to develop. I usually cherry pick the other commits then on the top of develop and reset the pr branch to that. Finally a force push of that branch into the remote repo is fixing it.

JKRhb:
I now rebased my branch to the latest state of `develop`, getting rid of the additional commit :)

JKRhb:
I think this PR should now be ready for review. Should I add a GitHub actions workflow for ensuring that tinydtls can be built using MingW and Visual Studio?

dtls_debug.c
```diff
@@ -252,7 +252,7 @@ dsrv_log(log_t level, const char *format, ...) {
   va_end(ap);
   fflush(log_fd);
 }
-#elif defined (HAVE_VPRINTF) /* WITH_CONTIKI */
+#elif defined (HAVE_VPRINTF) || defined(IS_WINDOWS) /* WITH_CONTIKI */
```
boaks, line 255: Does cmake not detect that on windows automatically?

At least, "AutoConf.cmake" contains `check_function_exists (vprintf        HAVE_VPRINTF)` which should to the job.

dtls_debug.h
```diff
@@ -24,6 +24,10 @@
 #include "global.h"
 #include "session.h"
 
+#ifdef IS_WINDOWS
```
boaks, line 27: Not sure, why this is required?

In "dtls_debug.c" it's simply included with:

```
#include <stdio.h>
```

If the the include is really required, I would go without `#ifdef IS_WINDOWS`.

dtls_debug.h
```diff
@@ -73,7 +77,7 @@ void dtls_set_log_level(log_t level);
  * Writes the given text to \c stdout. The text is output only when \p
  * level is below or equal to the log level that set by
  * set_log_level(). */
-#ifdef HAVE_VPRINTF
+#if defined(HAVE_VPRINTF) || defined(IS_WINDOWS)
```
boaks, line 80: See above, `HAVE_VPRINTF` should be auto-detected.

dtls_time.c
```diff
@@ -71,7 +136,7 @@ dtls_clock_init(void) {
 }
 
 void dtls_ticks(dtls_tick_t *t) {
-#ifdef HAVE_SYS_TIME_H
+#if defined(HAVE_SYS_TIME_H) || defined(IS_WINDOWS)
```
boaks, line 139: See above, `HAVE_SYS_TIME_H` should be auto-detected.

sha2/sha2.c
```diff
@@ -32,7 +32,12 @@
  * $Id: sha2.c,v 1.1 2001/11/08 00:01:51 adg Exp adg $
  */
 
+#if defined(_WIN32) || defined(_WIN64)
+#include "..\tinydtls.h"
```
boaks, line 36: Is this required?

`target_include_directories(tinydtls PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR})` should make it obsolete.

boaks:
Though I have no Windows system setup for such a build, I can't test it.
Some points are irritating, I considered, that `cmake` takes care of it. Not sure, if Windows is really that different.

boaks:
@JKRhb 

Maybe, if you can give us some more details about the used environment (install, etc.).
So that I can check on my own, if my comments apply.
I think, that would help.

Also, if your not able to longer work on this, please also inform us. Maybe I can plan to do the cleanup on my own and add that with an second commit.

JKRhb:
Thanks for the pinging and your review comments, @boaks! I think the PR should be cleaner now, relying on CMake where possible. There are still some places where workarounds for Visual Studio are necessary, I distinguished these cases using the `_MSC_VER` instead of the newly defined `IS_WINDOWS` macro now, removing unnecessary checks if possible. Let me know if you see any more spots where things could be improved.

Regarding testing, you should be able to compile for Windows under Linux using gcc and MingW (see my initial comment). I also tested the current state of the PR on a Windows machine using gcc and Visual Studio, and both variants worked. As mentioned above, building for the two environments could also be included in a GitHub Actions workflow, if you think this would make sense.

Lastly, I suppose it could also be useful to add a few lines regarding Windows support to the readme file?

dtls.c
```diff
@@ -1608,7 +1608,7 @@ dtls_0_send_hello_verify_request(dtls_context_t *ctx,
 			     dtls_ephemeral_peer_t *ephemeral_peer,
 			     uint8 *data, size_t data_length)
 {
-  uint8 buf[DTLS_RH_LENGTH + DTLS_HS_LENGTH + data_length];
+  uint8 buf[DTLS_RH_LENGTH + DTLS_HS_LENGTH + sizeof(uint16) + sizeof(uint8) + DTLS_COOKIE_LENGTH];
```
boaks, line 1611: Could you give some explanation about this change?

Why do you consider to replace the `data_length` by `sizeof(uint16) + sizeof(uint8) + DTLS_COOKIE_LENGTH`? Does your compiler requires constant values here?

JKRhb: > Does your compiler requires constant values here?

Yes, exactly, that was the reason for the change. The value for `data_length` is taken from here, therefore it still works:

https://github.com/eclipse/tinydtls/blob/bda40789a7c280f248eeca6d09ddd624cdaf5dc8/dtls.c#L1977-L1985

The current solution is, of course, very provisional. Would it make sense to turn `data_length` into a constant? Or do you see a better alternative?

boaks: @obgm 

FMPOV, if that is required for some compilers (maybe there are also compile switches to enable the feature?) then it seems to be reasonable. If we go for that, I think we should remove the obsolete parameter.

mrdeep1: Yes, in my experience, some compilers cannot handle dynamic stack allocation.

boaks: I check the code (and found it's mainly my changes for the stateless hello).

My proposal:

Replace `sizeof(uint16) + sizeof(uint8) + DTLS_COOKIE_LENGTH` by `DTLS_HV_LENGTH + DTLS_COOKIE_LENGTH`. Leave the parameter, because it contains the number of bytes to be copied from buf. Maybe an assert helps then to align the sizes of the copied bytes and `buf`.

(Before that stateless change, `dtls_send_multi` was used, which allocated the `sendbuf[DTLS_MAX_BUF]`. Maybe, in order to support `DTLS_CONSTRAINED_STACK`, we switch to that size/approach?)

JKRhb: > Replace sizeof(uint16) + sizeof(uint8) + DTLS_COOKIE_LENGTH by DTLS_HV_LENGTH + DTLS_COOKIE_LENGTH. Leave the parameter, because it contains the number of bytes to be copied from buf. Maybe an assert helps then to align the sizes of the copied bytes and buf.

Thanks, that looks good :) I applied your suggestion in a99838f830b71ae4f38fa6a20f97bf8e033ed52e.

> (Before that stateless change, dtls_send_multi was used, which allocated the sendbuf[DTLS_MAX_BUF]. Maybe, in order to support DTLS_CONSTRAINED_STACK, we switch to that size/approach?)

Do you mean that we should use the following instead, changing `buf` to `sendbuf`?

```c
#ifndef DTLS_CONSTRAINED_STACK
  unsigned char sendbuf[DTLS_MAX_BUF];
#endif /* ! DTLS_CONSTRAINED_STACK */
```

boaks: > Do you mean that we should use the following instead, changing buf to sendbuf?

Not with this PR. 

I also consider, to "allocate" the `buf` in `dtls_0_verify_peer`, I will check that later.

platform-specific/win-gettimeofday.c
```diff
@@ -0,0 +1,55 @@
+/* 
+ * Author: Ugo Varetto - ugovaretto@gmail.com
```
boaks, line 2: AFAIK, the name of the contributor is Jan Roman?
Who is Ugo Varetto?
If he is a co-author, either split his contribution into a different commit/PR, with "Signed-off-by: ugovaretto@gmail.com". And ugovaretto@gmail.com needs to be used to sign the ECA.

platform-specific/dtls_prng_win.c
```diff
@@ -0,0 +1,45 @@
+/*******************************************************************************
+ *
+ * Copyright (c) 2022 Olaf Bergmann (TZI) and others.
```
boaks, line 3: The common practice is to put your name here, if you created the file.
Other projects seems to use " Contributors to the Eclipse Foundation", but for now, I think, your name will be the right one.

platform-specific/win-gettimeofday.c
boaks: Sorry, I overseen that in my first review.

platform-specific/dtls_prng_win.c
JKRhb: Thank you, I adjusted the copyright statement in the latest fixup commit.

platform-specific/win-gettimeofday.c
JKRhb: The code by Ugo Varetto is actually third party code I included to allow for a reuse of the existing function implementation based on `gettimeofday()` in `dtls_time.c`. I noticed, however, that this is probably not needed and the Windows-specific time API can probably be used instead, resolving all legal matters in the process. Sorry for the inconvenience – I hope to be able to provide a rework soon.

boaks: Third-party work requires a proper license by that third party. So please, if possible, either ask the author for permission (best ECA by him) or try to redesign it.

platform-specific/dtls_prng_win.c
```diff
@@ -1,6 +1,6 @@
 /*******************************************************************************
  *
- * Copyright (c) 2022 Olaf Bergmann (TZI) and others.
+ * Copyright (c) 2022 Jan Romann.
```
boaks, line 3: + "and others".

JKRhb: Thanks! Should be resolved now :)

platform-specific/win-gettimeofday.c
JKRhb: I have now replaced the reimplementation of `gettimeofday` using a call of the Windows-specific `GetSystemTime` function. Things still work as before, getting rid of the additional code (and the potential legal issues).

dtls.h
```diff
@@ -324,8 +324,16 @@ void dtls_check_retransmit(dtls_context_t *context, clock_time_t *next);
 #define DTLS_CT_HANDSHAKE          22
 #define DTLS_CT_APPLICATION_DATA   23
 
+#ifdef __GNUC__
+#define PACK( __Declaration__ ) __Declaration__ __attribute__((__packed__))
+#endif
```
boaks, line 329: Not "mandatory", I would prefer a 

```
#elif defined(_MSC_VER)  
```

dtls.h
```diff
@@ -324,8 +324,16 @@ void dtls_check_retransmit(dtls_context_t *context, clock_time_t *next);
...
+
+#ifdef _MSC_VER
+#define PACK( __Declaration__ ) __pragma( pack(push, 1) ) __Declaration__ __pragma( pack(pop))
+#endif
+
```
boaks, line 334: Before this, the default, (may be a failing one) was `__attribute__((__packed__))`.
I would go for an explaining "#error", if PACK is not defined.

dtls_time.c
```diff
@@ -64,13 +69,13 @@ dtls_ticks(dtls_tick_t *t) {
...
 
-#elif defined(WITH_POSIX)
+#elif defined(WITH_POSIX) || defined(IS_WINDOWS)
 
 time_t dtls_clock_offset;
 
 void
 dtls_clock_init(void) {
-#ifdef HAVE_TIME_H
+#if defined(HAVE_TIME_H) || defined(_MSC_VER)
```
boaks, line 78: Is there a reason for mixing `IS_WINDOWS` `_MSC_VER`?

At least the `HAVE_TIME_H`, why doesn't this work for CMake in Windows?

boaks:
The left overs seems to be minor, so from my viewpoint, please "squash" and we will see, when @obgm can have a look at it.

obgm:
Yes, please squash for final review.

dtls.h
JKRhb: Should be addressed in 7126ffca98564345c05bb5cf6fb69e4b67784ac4, maybe the error message could be improved, though.

dtls_time.c
JKRhb: I misread the macro here as `HAVE_SYS_TIME_H`. As `HAVE_TIME_H` is also available on Windows/Visual Studio, this change can actually be reverted :) (Done in ae898f80932a6df6e24fe54054948ec7ae816282.)

boaks:
I miss the squashed commits.

boaks:
Maybe a misunderstanding on my side. I would put all the work in 1 commit, not 7.

Any arguments to keep in 7?

@obgm 

what do you prefer?

JKRhb:
Oh, I see! Squashing into one commit would be fine for me, alternatively this could also happen when the PR is merged (instead of creating a merge commit).

obgm:
please squash

JKRhb:
Squashed the seven commits into one now

obgm:
BTW: Flexible array members are allowed in C99 which is passed explicitly as `-std=c99`. A compiler that does not support this is not compliant. Its a surprise that a modern MS C-compiler has this flaw.

JKRhb:
> BTW: Flexible array members are allowed in C99 which is passed explicitly as `-std=c99`. A compiler that does not support this is not compliant. Its a surprise that a modern MS C-compiler has this flaw.

This is a good point. However, after doing some more research, I concluded that arrays of variable length are not supported by Visual Studio (see [here](https://social.msdn.microsoft.com/Forums/en-US/fbbf7664-8309-4599-93d1-8f83324ab155/visual-studio-wont-let-you-create-an-variable-sized-array-in-c?forum=vcgeneral), for example) in general. Therefore, setting the `C_STANDARD` CMake property to 99 (as specified [here](https://cmake.org/cmake/help/latest/prop_tgt/C_STANDARD.html)) does not have an effect and the compilation keeps failing with the following errors:

```
tinydtls\ecc\ecc.c(160,23): error C2057: expected constant expression
tinydtls\ecc\ecc.c(160,26): error C2466: cannot allocate an array of constant size 0
tinydtls\ecc\ecc.c(160,11): error C2133: 'temp': unknown size
```
So I suppose we would need to keep the current changes to also support Visual Studio. Using a static array length seems to be a better practice in general, though (at least according to [this article](https://www.phoronix.com/news/Linux-Kills-The-VLA) cited in the [Wikipedia article](https://en.wikipedia.org/wiki/Variable-length_array#C99) on VLAs), so I think this change can actually be considered an improvement.

> I miss some short instructions.
> E.g. Visual Studio, which extensions are used.
> Or if cygwin is supported, how to build it with that.

Sure, I'll add some instructions to the Readme file :)

JKRhb:
Added Windows-related instructions to the README in 21c137e, let me know if you see room for improvement here.

boaks:
> do you see anything missing for this PR to proceed? 

I currently a little too busy with [Eclipse/Californium](https://github.com/eclipse-californium/californium).
I hope, I can spend some time to finalize my part of the review next week.

boaks:
From my side LGTM.

obgm (commented):
Looks good so far, just needs the suggested assert.

ecc/ecc.c
```diff
@@ -157,7 +157,7 @@ static int fieldSub(const uint32_t *x, const uint32_t *y, const uint32_t *modulu
 //finite Field multiplication
 //32bit * 32bit = 64bit
 static int fieldMult(const uint32_t *x, const uint32_t *y, uint32_t *result, uint8_t length){
-	uint32_t temp[length * 2];
+	uint32_t temp[(arrayLength + 1) * 2];
```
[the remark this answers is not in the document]
JKRhb: Thank you for the suggestion, I updated the commit accodingly :)

boaks:
The assert was added.
LGTM?

mrdeep1:
The CMake build checks - in particular the Windows one, are throwing up lots of warnings that should be resolved.

boaks:
I have no windows compile system (nor do I plan to setup one.).

What I see in the logs of the github build , is

- "%zu" is not supported for printf.

```
D:/a/tinydtls/tinydtls/dtls.c: In function 'dtls_prepare_record':
D:/a/tinydtls/tinydtls/dtls.c:1403:16: warning: unknown conversion type character 'z' in format [-Wformat=]
 1403 |     dtls_alert("The sendbuf (%zu bytes) is too small\n", *rlen);
```

- and the random function complains about 

```
In file included from D:/a/tinydtls/tinydtls/dtls_prng.c:34:
D:/a/tinydtls/tinydtls/platform-specific/dtls_prng_win.c: In function 'dtls_prng':
D:/a/tinydtls/tinydtls/platform-specific/dtls_prng_win.c:34:11: warning: implicit declaration of function 'rand_s'; did you mean 'rand'? [-Wimplicit-function-declaration]
   34 |     err = rand_s(&number);
      |           ^~~~~~
      |           rand
D:/a/tinydtls/tinydtls/platform-specific/dtls_prng_win.c:34:11: warning: nested extern declaration of 'rand_s' [-Wnested-externs]
```

Is there more?
Anyway, @JKRhb  could you please try to fix them? do you have the same warnings building locally?

boaks:
Did you try to check the compiler options about the "%z" in printf?

JKRhb:
> Did you try to check the compiler options about the "%z" in printf?

I am currently doing a bit more research in this regard, it seems a bit tricky to get this to work without warnings in MinGW (although the compiled result works as expected). I'll try to resolve the issues as soon as possible, though.

In the meantime, however, I managed to get rid of the warnings regarding `rand_s` with d216467ae6ac70d716689a8f0f5f3b2ed6a46cd6.

mrdeep1:
I think  that there needs to be a way of forcing the github workflow builds to add in -Werror (gcc/clang) or -WX (Windows) options such that warnings become failures and easy to pick up.

JKRhb:
After a while spent on debugging, I found the reason for the `printf` warnings under MingW and was able to provide a fix :) Also provided a workaround for the remaining warnings that resulted from an empty function body in `dsrv_print_addr`, as the `inet_ntop` cannot be found under Windows. With the warnings resolved, I hope we are now ready to merge.

dtls_debug.h
```diff
@@ -78,11 +78,11 @@ void dtls_set_log_level(log_t level);
  * level is below or equal to the log level that set by
  * set_log_level(). */
 #ifdef HAVE_VPRINTF
-#if (defined(__GNUC__))
+#if (defined(__GNUC__) && !defined(__MINGW32__))
 void dsrv_log(log_t level, const char *format, ...) __attribute__ ((format(printf, 2, 3)));
 #else /* !__GNUC__ */
```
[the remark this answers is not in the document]
JKRhb: Thank you, good catch :)

boaks:
As far as I can see it, the warnings are resolved.

@mrdeep1 
Any open concerns from your side?

@obgm 
Your "final" conclusion?

Should we wait until the feedback? Should we merge it to develop?

mrdeep1:
There currently is an issue with github processing mac-os builds.  Looking at https://github.com/eclipse/tinydtls/actions/runs/3125449535/jobs/5069828382 I am not sure that this is building OK.

mrdeep1:
Agreed, this is not an issue related to this PR, but does need to be resolved.  I see that this also was not picked up in #166 - I need to find out why.

dtls_debug.h
```diff
@@ -78,11 +78,11 @@ void dtls_set_log_level(log_t level);
  * level is below or equal to the log level that set by
  * set_log_level(). */
 #ifdef HAVE_VPRINTF
-#if (defined(__GNUC__))
+#if (defined(__GNUC__) && !defined(__MINGW32__))
```
JKRhb, line 81: @mrdeep1 Could it be that (some of) the macOS related warnings also result from here? If I am not mistaken, clang also defines `__GNUC__`, so maybe we need to add `&& !defined(__clang__)` or `!defined(__llvm__)` here to get rid of the warnings.

JKRhb: I tried out adding `&& !defined(__clang__)` but it did not have any effect on the macOS warnings. I suppose this should be addressed in a follow-up PR then.

boaks: AFAIK, these MAC OS warnings are not related to the changes here in this PR.
These warnings occurs also on other PRs without the changes of this one.
So, no need to care.

obgm:
LGTM
@boaks please merge when you give it a go

mrdeep1:
I know how to fix the mac-os build errors and am working on it - and should have it ready in a few hours time.

boaks:
@mrdeep1 

But that mac-os build doesn't stop this PR, it's independent?
Ok, to merge it?

obgm:
I guess we can wait for a few hours then

mrdeep1:
Sorry - wrote comment in a rush as I had to go out.  I will create a separate PR for it, so thins one can be merged.

JKRhb:
Thank you for merging and for your reviews!

boaks:
Thanks for the patience.

PR #128, opened 2022-03-26
Rename condition WITH_RIOT_GNRC to WITH_RIOT_SOCK

HendrikVE:
Using tinydtls on RIOT OS is not limited to GNRC as network stack. It is also working with e.g. lwIP, see: https://github.com/RIOT-OS/RIOT/pull/17552. Therefore the name WITH_RIOT_GNRC is misleading.

PR in RIOT OS has already been approved and was merged: 
- https://github.com/RIOT-OS/RIOT/pull/17765


obgm:
Thanks!

PR #129, opened 2022-03-26
Add IPv4 support for RIOT OS

HendrikVE:
Add the capability to communicate using IPv4.

PR in RIOT OS has already been approved and was merged: 
- https://github.com/RIOT-OS/RIOT/pull/17765

Dependencies:
- #127
- #128

I'll rebase once the dependencies are merged.

Off-topic:
At first glance the project seemed to be stale, but after looking into the develop branch it became clear it is not. I just wanted to point out that some first-time users (like me) might get the wrong impression when they see that the latest commit to master is from July 2020 ;)

### DIFF
--- a/session.h
+++ b/session.h
@@ -32,14 +32,23 @@ typedef struct {
 } session_t;
  /* TODO: Add support for RIOT over sockets  */
 #elif defined(WITH_RIOT_SOCK)
+#include "net/ipv4/addr.h"
 #include "net/ipv6/addr.h"
 typedef struct {
-  unsigned char size;     /**< size of session_t::addr */
+  unsigned char size;       /**< size of session_t::addr */
   struct {
-    unsigned short port;  /**< transport layer port */
-    ipv6_addr_t addr6;    /**< IPv6 address */
-  } addr;                 /**< session IP address and port */
-  int ifindex;            /**< network interface index */
+    unsigned short family;  /**< IP address family */
+    unsigned short port;    /**< transport layer port */
+    union {
+#ifdef SOCK_HAS_IPV4
+      ipv4_addr_t ipv4;     /**< IPv4 address */
+#endif
+#ifdef SOCK_HAS_IPV6
+      ipv6_addr_t ipv6;     /**< IPv6 address */
+#endif
+    };
+  } addr;                   /**< session IP address and port */
+  int ifindex;              /**< network interface index */
 } session_t;
 #else /* ! WITH_CONTIKI && ! WITH_RIOT_SOCK */
 
--- a/session.c
+++ b/session.c
@@ -32,11 +32,23 @@
    && uip_ipaddr_cmp(&((A)->addr),&((B)->addr))			\
    && (A)->ifindex == (B)->ifindex)
 #elif defined(WITH_RIOT_SOCK)
-#define _dtls_address_equals_impl(A,B)                          \
-  ((A)->size == (B)->size                                       \
-   && (A)->addr.port == (B)->addr.port                                    \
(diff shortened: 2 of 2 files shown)


boaks:
In order to contribute to a Eclipse project, you need to sign a [ECA](https://www.eclipse.org/legal/ECA.php), see [CONTRIBUTING](https://github.com/eclipse/tinydtls/blob/master/CONTRIBUTING.md).

boaks:
Is that common in RIOT, to select IPv4 or IPv6 at compile time?
If not, I would consider a more generic solution, which decides that at runtime.
That would also make it easier, to use tinydtls as link-library.

HendrikVE:
> Is that common in RIOT, to select IPv4 or IPv6 at compile time?

Yes, RIOTs sock abstraction for instance will provide the IPv6 member conditionally at compile time, with all of its implications. At the moment the IPv4 member will be always present, so checking for `SOCK_HAS_IPV4`, which was just introduced recently, is not strictly necessary, but that may change later. 
In the end it is not an issue of availability of members in tinydtls: `ipv6_addr_t` and `ipv6_addr_equal` for example will be always available regardless the compile time options. However without compile time options, in the case of IPv4-only that would add 208 bytes to the `.text` section and for IPv6-only it would add 44 bytes without the need for it and I experienced a few times that the RIOT community is quite pedantic when it comes to memory usage.

boaks:
I'm not sure, if I get it. It sounds, that IPv6 must be enabled at compile time, but the other sentence seems to indicate, that IPv4 is always usable.

So, the configurations are:

- IPv4 (only)
- IPv4 and IPv6 (if enabled at compile time)

Or am I mislead?

HendrikVE:
I had a closer look on the codebase. As far as I can tell all important parts that would require IPv4 are left out conditionally by the compile-time switch `SOCK_HAS_IPV4`. The lines of codes that are not explicitly protected by it are covered by checking the family type for `AF_INET`. The confusing part is the fact that the IPv4 address member is within a union with IPv6, thus making no difference at this point, but its existence in a IPv6-only mode does not make sense and should also be removed. But these are things to be improved in RIOTs codebase.

So in summary we have all possible configurations:
- IPv4 -only
- IPv6-only
- IPv4 and IPv6 dualstack

obgm:
@HendrikVE can you please rebase?

obgm:
Looks good, ready to merge right after doing PR #127.

HendrikVE:
Rebased :)

(Sorry for force-pushing multiple times, I tried to use another email address of mine, but ECA failed with it)

session.h
```diff
@@ -32,14 +32,23 @@ typedef struct {
...
+    unsigned short family;  /**< IP address family */
+    unsigned short port;    /**< transport layer port */
+    union {
+#ifdef SOCK_HAS_IPV4
+      ipv4_addr_t ipv4;     /**< IPv4 address */
+#endif
+#ifdef SOCK_HAS_IPV6
+      ipv6_addr_t ipv6;     /**< IPv6 address */
+#endif
+    };
```
obgm, line 49: One potential issue I see here is that some (older) compilers do not allow anonymous unions. Unfortunately, I do not remember which one that was.

HendrikVE: I think this is less of an issue since this code is compiled only for RIOT OS and RIOT already uses anonymous unions. Mostly in driver code, but anonymous unions can also be found in parts of RIOT's primary network stack `gnrc`. So I would argue that compilers that do not allow anonymous unions are probably not considered anyway.

obgm: Sounds reasonable.

obgm (approved):

boaks:
merged to develop, cherry picked to main.

PR #132, opened 2022-05-12
CMake fixes

boaks:
Based on the initial PR #18 this PR adds support for getrandom and libcunit.
Also adds license headers and applies minor fixes.
Amend README.md with instructions to use cmake.


boaks:
@obgm 

Hi Olaf,

according your [comment](../18#issuecomment-1055315112)

> The major obstacle is that we would need someone who is willing to work on this, i.e., keep it in sync and functionally equivalent to the configure/make mechanism.

Generally it's not to hard to maintain cmake, especially, if it's marked as experimental.
Most header, function or field check have their equivalent in cmake. That stuff is isolated in "AutoConf.cmake" and "dtls_config.h.cmake.in".

Though it's also the base for a zephyr port, I would appreciate, if you review it, (optional) test it, and then merge it.

CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+
+set(PACKAGE_NAME "tinydtls")
+set(PACKAGE_VERSION "0.8.6" )
+
+option(DTLS_ECC "disable/enable support for TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8" ON )
+option(DTLS_PSK "disable/enable support for TLS_PSK_WITH_AES_128_CCM_8" ON)
+
+configure_file(dtls_config.h.cmake.in dtls_config.h )
+
+add_library(tinydtls)
```
JKRhb, line 44: For my use cases, I need tinydtls compiled to a shared/dynamically linked library like so:

```suggestion
add_library(tinydtls SHARED)
```

Could there be an option added that allows for enabling this mode of compilation?

JKRhb (commented):
I am currently trying out the CMake support in the context of #126 and noticed that adjustments for the `.gitignore` file are currently missing.

So far, I came up with the following:

```
# CMake

CMakeCache.txt
cmake_install.cmake
CMakeFiles
```

Do you think there is anything missing?

Another aspect concerns static vs. dynamic linking, see the comment below.

CMakeLists.txt
boaks: AFAIK, using

```
cmake -DBUILD_SHARED_LIBS=ON  ../tinydtls
```
will do that without the request change. I added this to the CMake section in the README.

If possible, please test, if that "works for you". 

(By the way, I need to test the install stuff, I skipped that in order to get faster feddbacks.)

boaks:
Add the new cmake temporary files to .gitignore.
(If a separate build directory is used, these temporary files will also be outside of git working tree.)

boaks:
I've tested the install and extended it with the version for shared libraries.

CMakeLists.txt
JKRhb: I just tested it and generating the shared library works fine :) However, for some reason there are three files which are being generated:

```
libtinydtls.so (symbolic link to libtinydtls.so.0)
libtinydtls.so.0 (symbolic link to libtinydtls.so.0.8.6)
libtinydtls.so.0.8.6
```

Using `libtinydtls.so.0.8.6` (after renaming it) works as intended, however, it would probably be better if only `libtinydtls.so` was generated as an actual file?

JKRhb: > Ah, that's the way, unix does versioning for shared libraries with symbolic-links. Should I add a option to remove that?

Oh, sorry, I saw you comment a bit too late :/ Yeah, that would be great :)

boaks: Yes, as I wrote above, I added that in order to do it in the common way for unix. At least I consider that as the common way. I can add an option to suppress that.

boaks: Let me ask: beside that you expected to have only one file, do you consider any downsides?

boaks: My question is:
You expect one file, ok.
You get three files, including the version pattern common in unix.
What is the downside in your opinion in having three instead of one file?

JKRhb: You are right, there is actually no real reason to change it to one file, copying over all three works fine :) Sorry for the delay, and thank you for your patience!

boaks: You're welcome.

obgm: Excellent, I guess this is resolved then.

obgm:
Thanks, @JimmyBjorklund  @boaks and @JKRhb for making this happen.

PR #139, opened 2022-05-25
Fix issue 134

boaks:


bathooman:
Based on my tests, The PR indeed resolves the bug.

obgm (commented):
LGTM. I suggest to remove the dead code enclosed in `#if 0 ... #endif`.

boaks:
Great!

I will do the house-keeping (remove the dead code) and cherry-pick it to "main" at the weekend or on Monday.
That completes the process switch ;-). I think creating a temporary develop-archive branch, for those who want to verify, that nothing is lost on their own. And then a new "develop" based on the current "main".
I'm looking forward to have that cleaned up.

PR #141, opened 2022-05-25
dtls.c: consider data_length processing incoming alert.

boaks:
Ignore malformed alerts instead of process them as fatal_alert.


bathooman:
Based on my tests, The PR indeed resolves the bug.

obgm:
LGTM

Issue #143, opened 2022-05-26
Issues for general cleanup?


boaks:
As mentioned in [add ccm - comment on too many whitespace change](https://github.com/eclipse/tinydtls/pull/105#issuecomment-1138191402) the current code seems the have a mixed up format.

Therefore I would like to collect some input for a general cleanup.

From my side:
- whitespace format
- license update to EPL v2.0
- logging: 
    - check messages (e.g. calls `dtls_debug_dump` already with a ":" at the end of the name),
    - remove eols ("\n") from the messages and add that to the print functions in `dtls_debug`.

Such general cleanups usually "locks the code for some time", so maybe it's also worth to get some feedback, what should be done before (e.g. merging selected open PRs).

obgm:
One task would be to adjust existing code to follow the basic coding style outlined in  [CONTRIBUTING](https://github.com/eclipse/tinydtls/blob/main/CONTRIBUTING.md) and enforce contributions to follow that style. In another project, I am currently testing a `clang-format` specification that does almost the right thing. (The problem with these tools is that you need to write a lot of annotations where you want to deviate from the strict formatting rules.)

obgm:
Regarding license update: What would be the reasons for undergoing the enormous pain of changing the license model? I presume you would not do this unless there is something entirely wrong with the existing EPL v1.0/EDL v1.0.

obgm:
> I may be wrong, but I assume, a new release requires a 2.0.

Okay, more reading to do for release preparation.

The pain arises from reading legal stuff: All tinydtls users will have to make their lawyers process the new license. It is not the editing that makes license changes difficult but the changed contents of the license. I have not yes compared the two documents or read discussions that may have commented on the changes but then again, this is exactly what I meant with "pain".

boaks:
[EPL-v1.0-FAQ](https://www.eclipse.org/legal/eplfaq.php)

[EPL-v2.0-FAQ](https://www.eclipse.org/legal/epl-2.0/faq.php)

[Diffs](https://www.eclipse.org/legal/epl-2.0/faq.php#h.a0eux401qus)

Yes, some may require to read it.
But many people and companies are common to EPL 2.0  and therefore I consider, that processing it by lawyers is a rare exception. 
EPL 2.0 is not intended to make the use harder, it's intended to make it easier.

boaks:
[...]

PR #147, opened 2022-06-23
Add CCM cipher suite variants (2.)

boaks:
In case of CCM gets recommended and CCM8 deprecated.

Introduce cipher_suite_param_t for cipher-suite specific mac_len and
key_exchange_algorithm, and introduce dtls_cipher_index_t for simplified
cipher-suite parameter lookup. Cleanup old functions.


crypto.h
```diff
@@ -105,7 +111,7 @@ typedef struct {
 typedef struct {
   dtls_compression_t compression;	/**< compression method */
 
-  dtls_cipher_t cipher;		/**< cipher type */
+  dtls_cipher_index_t cipher_index;	/**< internal index for cipher_suite_params, DTLS_CIPHER_INDEX_NULL for TLS_NULL_WITH_NULL_NULL */
```
mrdeep1, line 114: Line length is > 80 chars, breaking coding style.  This occurs in multiple places

dtls.c
```diff
@@ -2801,7 +2866,7 @@ dtls_send_client_hello(dtls_context_t *ctx, dtls_peer_t *peer,
   psk = is_psk_supported(ctx);
   ecdsa = is_ecdsa_supported(ctx, 1);
 
-  cipher_size = 2 + ((ecdsa) ? 2 : 0) + ((psk) ? 2 : 0);
+  cipher_size = 2 + ((ecdsa) ? 4 : 0) + ((psk) ? 4 : 0);
```
mrdeep1, line 2869: It would be good to have a comment here that the `4` refers to there being 2 algorithms that are going to be defined - to make it more obvious what needs to be done when another algorithm is added.
Perhaps `4` should be `2*2` to emphasize this further.

mrdeep1 (commented):
Had a good read through the changes, only a couple of minor things found.

dtls.c
boaks: I tried to improve the "length documentation" in a commit later 84f834a69eefefd4180edcda5a1b485cadb75b07 .
I will add also comment here.

crypto.h
boaks: Should we break / split line for macros as well?

mrdeep1: That would be my preference. @obgm ?

dtls.c
mrdeep1: https://github.com/eclipse/tinydtls/commit/84f834a69eefefd4180edcda5a1b485cadb75b07 certainly helps, but does not specifically bring focus to this line for the supported suites that are going to be announced.
````
-  cipher_size = 2 + ((ecdsa) ? 2 : 0) + ((psk) ? 2 : 0);
+  /* Each announced cipher suite is 2 bytes long */
+  cipher_size = 2 + ((ecdsa) ? 2*2 : 0) + ((psk) ? 2*2 : 0);
````

crypto.h
boaks: That requires a lot of concentration to only fix the lines, which are altered with this PR.
Otherwise we get a couple of "format only changes", which I would prefer to have them done "once applying a formatter" (issue #143).
We will see ...

boaks: OK, if I continue to try to fix the format only of the changed lines, that will eat up hour and hour. it requires also to spend a lot of time in the follow-up commits, though these are changing the same lines and cherrypick ends up in also a lot of conflicts.
My proposal: I go "fast over it". And the final "format" approach is then done with #143.

mrdeep1: Fair enough.  Tidy up things with clang-format etc. at  a later stage.

dtls.c
boaks: Yes, that commit doesn't address it, because this lines are removed on a [commit before](https://github.com/eclipse/tinydtls/commit/cb16f58ea036fa971d9e4eb91c67b9b4d4219552#diff-947f252e5a6242133f45e2963433822253886eb9a6bf4c12c7163ac8e90fe017R2906).

Therefore I think, it doesn't really pay-off to polish it more and then remove it.

obgm (approved):

boaks:
Available on "develop" until follow up PR #148 gets merged.

boaks:
Now merged in main.

PR #149, opened 2022-06-25
Add client dtls connection ID (2.).

boaks:
Simple client side implementation indicates support and uses the cid of
the server, when negotiated by that.


mrdeep1:
It would be good if a server side API could be added where the calling application can define the CID to use so that it can be tracked by the application to change the tuple to send back any response on.

boaks:
> server side API

As for now, tinydtls CID only supports the client side. To be more precise, the client only indicates to support CID, but doesn't use a CID on it's own side. For the next time I will not able to spend the required time to implement a server side. If that changes, the proposed API will make a lot of sense.

PR #150, opened 2022-06-27
Use EPL-1.0 for recently new files.

boaks:
Currently tinydtls uses EPL-1.0 and mixing with EPl-2.0 is not
recommended. Therefore adjust the EPL-2.0 of new files to
EPL-1.0.


obgm:
Thanks for fixing this.

Issue #133, opened 2022-05-24
We reported a use-after-free bug via email, but have not gotten any responses yet.


nopitydays:
Hi,

We reported a use-after-free bug by sending an email to security@eclipse.org over a month ago.
However, we have not gotten any responses yet.

Any suggestions for reporting security bugs of the tinydtls project?

Thanks,
nop

boaks:
At least for now, the paper work to became a committer in tinydtls is not done.
Therefore I have no access to the security related issues of this project. 
I can just open a ticket and ask, if your e-mail was processed and converted into a security ticket.
Anyway, is it related to "master" or "develop"?

nopitydays:
> At least for now, the paper work to became a committer in tinydtls is not done. Therefore I have no access to the security related issues of this project. I can just open a ticket and ask, if your e-mail was processed and converted into a security ticket. Anyway, is it related to "master" or "develop"?

The bug reported a month ago is related to the "master" branch, and we just reported another bug that is related to the "develop" branch last night by sending an email to security@eclipse.org.

Thanks!

boaks:
> The bug reported a month ago is related to the "master" branch

Is it already fixed in the "develop" branch?
Or is the new e-mail about the same in that branch?
Anyway, I guess, I personally will not be able to see that information before mid of next week.

nopitydays:
> Is it already fixed in the "develop" branch?
Sorry for my unclear description.
The bug related to the "master" branch is fixed in the "develop" branch (the vulnerable code seems to have been refactored).
The bug related to the "develop" branch has not been fixed, and we can still reproduce it.

> Anyway, I guess, I personally will not be able to see that information before mid of next week.
Should we wait for email replies, or could we just open the bugs as the Github issues?

boaks:
> Should we wait for email replies, or could we just open the bugs as the Github issues?

That depends ...

If it's easy to be misused for attacks, I think the e-mail to [security@eclipse.org](mailto:security@eclipse.org) is the right thing. If it's just a nasty bug, and the probability, that it is misused for an attack is small, then maybe an issue.

Anyway, patience is always good for open source.
Please keep in mind, that many invest their free time in such projects. 
And at least I spend my free time also in other topics ;-).

nopitydays:
Agree :), we will keep waiting. Thanks for your suggestions!

mbarbero:
@nopitydays I've forwarded your vulnerability report to project lead. He should get back to you.

[...]

PR #152, opened 2022-07-06
dtls.c: fix off-by-one error in dtls_asn1_len()

obgm:
For ASN.1 integers greater than 128 (i.e., with the most significant bit set) the encoded length might exceed *data_len. The length check in the while loop must be done after decrementing *data_len.

This PR addresses a bug reported by Shisong Qin, cf. #133

The function can be tested like this:

```c
int main() {
  uint8_t input[] = { 0x87, 0xab, 0xcd };
  uint8_t *data = input;
  size_t n = sizeof(input);
  dtls_asn1_len(&data, &n);
}
```

Compile with 

```
clang -O0 -fsanitize=address  check_asn1_len.c -o check_asn1_len
```

The erroneous function would cause the sanitizer to report a stack-buffer-overflow (see below) while staying silent for the fixed version.

```
=21072==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffd6d54a863 at pc 0x0000004ce2d7 bp 0x7ffd6d54a740 sp 0x7ffd6d54a738
READ of size 1 at 0x7ffd6d54a863 thread T0
...
```

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -2005,11 +2005,9 @@ dtls_asn1_len(uint8 **data, size_t *data_len)
     (*data_len)--;
     if (octets && *data_len == 0)
       return (size_t)-1;
-    while (octets) {
+    while (octets > 0 && *data_len > 0) {
       len = (len << 8) + (**data);
       (*data)++;
-      if (*data_len == 0)
-        return (size_t)-1;
       (*data_len)--;
       octets--;
     }
--- a/dtls.c
+++ b/dtls.c
@@ -2003,9 +2003,9 @@ dtls_asn1_len(uint8 **data, size_t *data_len)
     size_t octets = (**data) & 0x7f;
     (*data)++;
     (*data_len)--;
-    if (octets && *data_len == 0)
+    if (octets > *data_len)
       return (size_t)-1;
-    while (octets > 0 && *data_len > 0) {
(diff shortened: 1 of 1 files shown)


boaks:
I would prefer to have PRs against "main".
Only if the review needs to be postponed, I would cherry-pick them on develop and mark that with a label in the PR.

dtls.c
```diff
@@ -2005,11 +2005,9 @@ dtls_asn1_len(uint8 **data, size_t *data_len)
     (*data_len)--;
     if (octets && *data_len == 0)
```
boaks, line 2006: If the issue is an invalid encoded ASN.1 length, (in the example it indicates a length of 7 bytes, but only 2 are available), an alternative solution maybe to check this ahead:

```
if (octets > *data_len)
```

obgm: Yes, an upfront check is even better. I will update the code in a minute.

boaks: And please, against "main" :-).

obgm: New code pushed as fixup for later squash.

boaks:
What's about using "main" instead of "develop"?

if the last changes are put in a PR against main, the LGTM and we can merge them fast into main.

obgm:
Sure, will do.

PR #153, opened 2022-07-06
dtls.c: fix off-by-one error in dtls_asn1_len()

obgm:
For ASN.1 integers greater than 128 (i.e., with the most significant bit set) the encoded length might exceed *data_len. The length check is done upfront to ensure that the specified number of octets does
not exceed the actual *data_len.

This bug was reported by Shisong Qin.

This PR supersedes PR #152.


obgm:
Thanks, here we go

PR #154, opened 2022-07-06
dtls.c: Fix incomplete release of peer

obgm:
In handle_0_verified_client_hello(), an existing peer is released and replaced by a new peer object. To ensure that references to this peer are cleared, dtls_destroy_peer() must be called with the unlink flag set.

Moreover, when handle_verified_client_hello() fails for the new peer structure, dtls_destroy_peer() must be called for that peer as well to ensure removal from the sendqueue.

This issue has been reported by Shisong Qin, cf. #133

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -3884,9 +3884,7 @@ handle_0_verified_client_hello(dtls_context_t *ctx, dtls_ephemeral_peer_t *ephem
   dtls_peer_t *peer = dtls_get_peer(ctx, ephemeral_peer->session);
   if (peer) {
      dtls_debug("removing the peer, new handshake\n");
-     DEL_PEER(ctx->peers, peer);
-
-     dtls_free_peer(peer);
+     dtls_destroy_peer(ctx, peer, 0);
      peer = NULL;
   }
   dtls_debug("creating new peer\n");
@@ -3928,8 +3926,7 @@ handle_0_verified_client_hello(dtls_context_t *ctx, dtls_ephemeral_peer_t *ephem
 
   err = handle_verified_client_hello(ctx, peer, data, data_length);
   if (err < 0) {
-    DEL_PEER(ctx->peers, peer);
-    dtls_free_peer(peer);
+    dtls_destroy_peer(ctx, peer, DTLS_DESTROY_CLOSE);
     return err;
   }
 


boaks:
```
static void dtls_destroy_peer(dtls_context_t *ctx, dtls_peer_t *peer, int unlink)
{
  if (peer->state != DTLS_STATE_CLOSED && peer->state != DTLS_STATE_CLOSING) {
    dtls_close(ctx, &peer->session);
  }
  if (unlink) {
    DEL_PEER(ctx->peers, peer);
    dtls_dsrv_log_addr(DTLS_LOG_DEBUG, "removed peer", &peer->session);
  }
  dtls_free_peer(peer);
}
```

From a first look, I don't think, that sending a "CLOSE_NOTIFY" is always a good idea. 
In a case, where the other peer seems to have closed it anyway.
Do I oversee something?

The part:
```
DEL_PEER(ctx->peers, peer);
...
dtls_free_peer(peer);
```

is already done, so the missing stuff seems to be the

```
```

Currently there is no call without 1 for "unlink". 
What was the idea behind that?

I would prefer to extend it to:

```
static void dtls_destroy_peer(dtls_context_t *ctx, dtls_peer_t *peer, int unlink, int close)
```

(or just rename the "unlink" in "close")

obgm:
I preferred the call to `dtls_destroy_peer()` here to avoid too much code replication. My concern was that at some point additional cleanup tasks might get added to `dtls_destroy_peer()` without updating the parts of the code that replicate (part of) this function's behavior. I am also not sure about closing off the session but as we are going to replace it, it might be a good idea to do so.

Regarding `unlink`: This is a good catch. At some point, there was a call with this flag cleared to avoid a double free. This has been replaced a while ago, so this parameter can be dropped AFAICT.

obgm:
This change is now rebased to use PR #155 to indicate that no CloseNotify is to be sent in this case.

dtls.c
```diff
@@ -3879,9 +3886,7 @@ handle_0_verified_client_hello(dtls_context_t *ctx, dtls_ephemeral_peer_t *ephem
   dtls_peer_t *peer = dtls_get_peer(ctx, ephemeral_peer->session);
   if (peer) {
      dtls_debug("removing the peer, new handshake\n");
-     DEL_PEER(ctx->peers, peer);
-
-     dtls_free_peer(peer);
+     dtls_destroy_peer(ctx, peer, 0);
```
boaks, line 3889: Here the 0 is right, don't close the previous session.

dtls.c
```diff
@@ -3923,8 +3928,7 @@ handle_0_verified_client_hello(dtls_context_t *ctx, dtls_ephemeral_peer_t *ephem
 
   err = handle_verified_client_hello(ctx, peer, data, data_length);
   if (err < 0) {
-    DEL_PEER(ctx->peers, peer);
-    dtls_free_peer(peer);
+    dtls_destroy_peer(ctx, peer, 0);
```
boaks, line 3931: At that point I guess, it's better to DTLS_DESTROY_CLOSE.
That indicates the other peer, that this peer doesn't continue (for whatever reason).

obgm: Thanks, I was unsure about this one. I have changed this accordingly.

boaks:
LGTM

obgm:
Here we go!

PR #155, opened 2022-07-06
dtls.c: introduce flags parameter for dtls_destroy_peer()

obgm:
The parameter unlink for dtls_destroy_peer() has been in use for several years and therefore can be safely removed as suggested in PR #154.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -62,6 +62,13 @@
 
 #define DTLS10_VERSION 0xfeff
 
+/* Flags for dtls_destroy_peer()
+ *
+ *  DTLS_DESTROY_CLOSE indicates that the connection should be closed
+ *                     when applicable
+ */
+#define DTLS_DESTROY_CLOSE 0x02
+
 #ifdef RIOT_VERSION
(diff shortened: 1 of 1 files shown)


boaks:
I would prefer to "reuse" it for the close.

obgm:
The fixup now introduces a flag to control sending of CloseNotify when applicable (i.e., not if the peer state is CLOSED or CLOSING).

dtls.c
```diff
@@ -62,6 +62,13 @@
 
 #define DTLS10_VERSION 0xfeff
 
+/* Flags for dtls_destroy_peer()
+ *
+ *  DTLS_DESTROY_CLOSE indicates that the connection should be closed
+ *                     when applicable
+ */
+#define DTLS_DESTROY_CLOSE 0x02
```
boaks, line 70: That's a nice idea.

In a future PR, I would then go to add

```
DTLS_DESTROY_WITHOUT_CLOSE 0x01
```

obgm:
Great, the squashed version now goes into main.

PR #158, opened 2022-07-22
dtls_debug: Add in support for separate logging handler

mrdeep1:
Allow the application to define a log handler callback to use to provide
consistency in the logging output logic.

### DIFF
--- a/dtls_debug.c
+++ b/dtls_debug.c
@@ -41,6 +41,7 @@ typedef int in_port_t;
 
 #include "global.h"
 #include "dtls_debug.h"
+#include "dtls_mutex.h"
 
 #ifndef min
 #define min(a,b) ((a) < (b) ? (a) : (b))
@@ -242,16 +243,11 @@ dsrv_print_addr(const session_t *addr, char *buf, size_t len) {
 #endif /* NDEBUG */
 
 #if !defined(WITH_CONTIKI) && !defined(_MSC_VER)
-void
-dsrv_log(log_t level, const char *format, ...) {
+
+static void
+dtls_logging_handler(log_t level, const char *message) {
   static char timebuf[32];
-  va_list ap;
-  FILE *log_fd;
-
-  if (maxlog < (int) level)
-    return;
-
-  log_fd = level <= DTLS_LOG_CRIT ? stderr : stdout;
+  FILE* log_fd = level <= DTLS_LOG_CRIT ? stderr : stdout;
 
   if (print_timestamp(timebuf,sizeof(timebuf), time(NULL)))
     fprintf(log_fd, "%s ", timebuf);
@@ -259,11 +255,62 @@ dsrv_log(log_t level, const char *format, ...) {
   if (level <= DTLS_LOG_DEBUG)
     fprintf(log_fd, "%s ", loglevels[level]);
 
-  va_start(ap, format);
-  vfprintf(log_fd, format, ap);
-  va_end(ap);
+  fwrite(message, strlen(message), 1, log_fd);
   fflush(log_fd);
 }
+
+static dtls_log_handler_t log_handler = dtls_logging_handler;
+
+void
+dtls_set_log_handler(dtls_log_handler_t app_handler) {
+  if (app_handler == NULL)
+    log_handler = dtls_logging_handler;
+  else
+    log_handler = app_handler;
+}
+
+#ifndef DTLS_DEBUG_BUF_SIZE
+#define DTLS_DEBUG_BUF_SIZE 128
+#endif
+
+#ifdef DTLS_CONSTRAINED_STACK
+static dtls_mutex_t static_mutex = DTLS_MUTEX_INITIALIZER;
+static char message[DTLS_DEBUG_BUF_SIZE];
+#endif /* DTLS_CONSTRAINED_STACK */
+
+/*
+ * Caution. If DTLS_CONSTRAINED_STACK is set, then the same mutex will get
+ * locked in dtls_log() and/or dtls_dsrv_hexdump_log() so these functions
+ * cannot call each other.  Furthermore, if log_handler() calls dsrv_log() /
+ * dtls_dsrv_hexdump_log() there will be a recursive lookup.
+ */
+void
+dsrv_log(log_t level, const char *format, ...) {
+  va_list ap;
+  size_t len;
+#ifndef DTLS_CONSTRAINED_STACK
+  char message[DTLS_DEBUG_BUF_SIZE];
+#endif /* ! DTLS_CONSTRAINED_STACK */
+
+  if (maxlog < (int) level)
+    return;
+
+#ifdef DTLS_CONSTRAINED_STACK
+  dtls_mutex_lock(&static_mutex);
+#endif /* DTLS_CONSTRAINED_STACK */
+  va_start(ap, format);
+  len = vsnprintf( message, sizeof(message), format, ap);
+  va_end(ap);
+  if (len + 1 > sizeof(message)) {
+    /* 6 is needed as trailing 0 byte space needed */
+    snprintf(&message[sizeof(message)-6], 6, " ...\n");
+  }
+  log_handler(level, message);
+#ifdef DTLS_CONSTRAINED_STACK
+  dtls_mutex_unlock(&static_mutex);
+#endif /* DTLS_CONSTRAINED_STACK */
+}
+
 #elif defined (HAVE_VPRINTF) /* WITH_CONTIKI */
 void
 dsrv_log(log_t level, char *format, ...) {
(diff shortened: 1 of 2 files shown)


dtls_debug.c
```diff
@@ -352,6 +374,43 @@ dtls_dsrv_hexdump_log(log_t level, const char *name, const unsigned char *buf, s
   if (maxlog < (int) level)
     return;
 
+  if (log_handler) {
```
boaks, line 377: Hm, maybe, using a default handler would obsolete also some code then on other places.
Did you already thought of something similar?

mrdeep1: Not really (did not want the code to do `print_timestamp()` for what is passed to the alternative handler), but now you raise it, it makes sense and I can see how to do it.  I will have a look at reducing the code duplication.

boaks: Alternatively, maybe we add a "compile switch" for those, who don't use it.

mrdeep1: Just pushed with code changed.  Not done the Contiki alternative as I have nothing to test code against.

dtls_debug.h
```diff
@@ -73,6 +73,23 @@ log_t dtls_get_log_level(void);
...
+/**
+ * Logging callback handler definition.
+ *
+ * @param level One of the DTLS_LOG_* values.
+ * @param message Zero-terminated string message to log.
+ */
+typedef void (*dtls_log_handler_t) (log_t level, const char *message);
+
+/* So applications using TinyDTLS know that this handler is available */
+#define HAVE_DTLS_SET_LOG_HANDLER
```
obgm, line 85: Hm, wouldn't applications' configure just check for the existence of `dtls_set_log_handler()` in the library and then set whatever symbol they want? I am not sure that we want to maintain this preprocessor symbol all the time.

dtls_debug.c
```diff
@@ -241,28 +241,51 @@ dsrv_print_addr(const session_t *addr, char *buf, size_t len) {
...
+}
+
+#ifndef DTLS_DEBUG_BUF_SIZE
+#define DTLS_DEBUG_BUF_SIZE 128
+#endif
+
+void
+dsrv_log(log_t level, const char *format, ...) {
+  va_list ap;
+  char message[DTLS_DEBUG_BUF_SIZE];
```
obgm, line 277: Since we have `dtls_mutex.h` we could also provide a version with a static buffer here.

dtls_debug.h
mrdeep1: It could, but it cannot test for the presence if the library is yet to be built as, say for a submodule. Another way would be to check against the version number of the TinyDTLS library so I went for the easy option.

dtls_debug.c
mrdeep1: Yes, will update it shortly.

mrdeep1: Code pushed

boaks:
Maybe a late and silly question:

If it's mainly about the timestamp, would a "header_callback" with level and optional timestamp do that job easier?
Or adding a compile switch with e.g. "LOGGING_SIMPLE_TIME" (or whatever name)?

mrdeep1:
Any additional function has the same set of challenges as finding this new logging function.

I'm just trying to find an easy way for determining availability of functionality.

- libcoap's configure can find if a function exists in externally provided tinydtls .h headers and/or tinydtls.{so|dll} and build accordingly
- libcoap can dynamically load libtinydtls.{so|dll}, check for the presence of a function and then use it if available
- libcoap can have tinydtls as a submodule which could look for the availability of a function - the easiest way to do this was to have HAVE_DTLS_SET_LOG_HANDLER.
- tinydtls makes use of different version numbers as functionality evolves.

Whatever the determination method is, libcoap and hence higher level applications need to be able to control how logging is done.

boaks:
Thanks!

maybe I over simplified it by assuming, that it is anyway "compiled in".
Sure, for a pre-compiled  library, it doesn't work to have just a compile switch.

dtls_debug.c
boaks: I would welcome to add a short comment about calling the handler with the mutex locked.

This PR doesn't change that, but also the `static char timebuf[32];` requires such protection. Maybe that gets fixed later in an additional PR.

dtls_debug.c
```diff
@@ -345,46 +382,51 @@ void dtls_dsrv_log_addr(log_t level, const char *name, const session_t *addr)
...
   if (extend) {
-    fprintf(log_fd, "%s: (%zu bytes):\n", name, length);
-
+    snprintf(message, sizeof(message), "%s: (%zu bytes):\n", name, length);
     while (length--) {
-      if (n % 16 == 0)
-        fprintf(log_fd, "%08X ", n);
-
-      fprintf(log_fd, "%02X ", *buf++);
+      len = strlen(message);
```
boaks, line 400: If the hex dump message, including newlines, is assembled  ahead before calling the handler, it my frequently truncate the output.
Should that be indicated by "..." at the end of the output?
Or should the handler be called per line (maybe adding an additional parameter for follow up lines)?

boaks:
A general question only weakly linked to this PR:

Currently the logging adds a "\n"  before the handler is called. For some OS  logging systems (e.g. zephyr) this is an additional one. I would therefore go to remove all trailing "\n" and add them in the handler in an additional PR.
WDYT?

dtls_debug.c
```diff
@@ -241,28 +242,64 @@ dsrv_print_addr(const session_t *addr, char *buf, size_t len) {
...
 
+  fwrite(message, strlen(message), 1, log_fd);
+  fflush(log_fd);
+}
+
+static dtls_log_handler_t log_handler = dtls_logging_handler;
+
+void
+dtls_set_log_handler(dtls_log_handler_t app_handler) {
+  if (log_handler == NULL)
```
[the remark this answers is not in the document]
mrdeep1: Good catch - fixed.

dtls_debug.c
mrdeep1: The external logging handler may also be truncating length - which may even affect a line by line mode.  My thoughts here that in the general case it does not really matter if information is lost, bu if space is needed to give the detail in the longer lines, then DTLS_DEBUG_BUF_SIZE can be redefined.  Furthermore, if `extend `is set as a parameter, then the information is not output over multiple lines.

I do like the concept though of "..." indicating something is missing. Code updated and pushed.

dtls_debug.c
mrdeep1: Actually,` dsrv_log() `should NOT be called with 'static_mutex' locked as it does that within the function.  I agree that` timebu[32]` ought to be protected.

boaks: My point about adding a comment is more, that if this callback locks an other mutex, and also someone calls the logging with that mutex locked, then this will be in danger of a deadlock. Therefore I think to add a comment, that informs about the mutex helps to prevent from such an deadlock.

mrdeep1: Added comment and code pushed.

dtls_debug.h
mrdeep1: Removed the need for HAVE_DTLS_SET_LOG_HANDLER to be defined and code pushed.

dtls_debug.h
```diff
@@ -73,6 +73,21 @@ log_t dtls_get_log_level(void);
 /** Sets the log level to the specified value. */
 void dtls_set_log_level(log_t level);
 
+/**
+ * Logging callback handler definition.
+ *
+ * @param level One of the DTLS_LOG_* values.
+ * @param message Zero-terminated string message to log.
+ */
+typedef void (*dtls_log_handler_t) (log_t level, const char *message);
```
boaks, line 82: I would prefer, if a comment about DTLS_DEBUG_BUF_SIZE is provided here.

mrdeep1: Comment added and code pushed.

dtls_debug.c
obgm: Instead of calling `strlen()` multiple times, I would prefer to calculate the available space from the return value of `snprintf()`. (Constant instead of linear complexity).

obgm (commented):
Just one small detail, otherwise, this looks good.

mrdeep1:
Unfortunately, as per the man page
````
The functions snprintf() and vsnprintf() do not write more than size bytes (including the terminating null byte ('\0')). If the output was truncated due to this limit then the return value is the number of characters (excluding the terminating null byte) which would have been written to the final string if enough space had been available. Thus, a return value of size or more means that the output was truncated. (See also below under NOTES.)
````
the returned size can be larger than the actual buffer size which is somewhat frustrating.

**Comment by boaks** [comment]
FMPOV, if the sum of the returns (len) exceeds the message buffer size, then the loop can be finished and we know it exceeds. 
Or did I oversee something?

**Comment by mrdeep1** [comment]
I have gone with @boaks suggestion, but I needed to check `len` overflow in multiple places.  Code pushed.

File: dtls_debug.c [hunk 7]
```diff
@@ -346,46 +389,56 @@ void dtls_dsrv_log_addr(log_t level, const char *name, const session_t *addr)
...
+    len = snprintf(message, sizeof(message), "%s: (%zu bytes): ", name, length);
+    while (length-- && len + 1 < sizeof(message)) {
+      len += snprintf(&message[len], sizeof(message)-len, "%02X ", *buf++);
+    }
   }
-  fprintf(log_fd, "\n");
-
-  fflush(log_fd);
+  /* len could be oversize, so re-calculate */
+  len = strlen(message);
```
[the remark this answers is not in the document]
mrdeep1: `len `at this point can be bigger than `sizeof(message)` because of previous `snprintf()`s that needed more space to write.

If `len `is untouched at this point, then random memory will get overwritten in the following lines if `message `is not big enough.  All very messy and prone to coding errors when relying on the returned length from `snprintf()`.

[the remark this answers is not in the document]
mrdeep1:
```
Hm, may be that discovered a hidden issue, the code replaces the 0 in the message buffer by \n. If that the
last byte in the buffer, that smells for a un-terminated string.
```
`snprintf()` is not allowed to do that if there is insufficient space.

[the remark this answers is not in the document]
boaks:
```
if (len + 2 < sizeof(message))
   snprintf(&message[len], sizeof(message)-len, "\n");
else
   snprintf(&message[sizeof(message)-5], 5, " ...\n");
```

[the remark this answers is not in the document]
mrdeep1: Yes, we can reverse the order of doing the `snprintf()` by changing to your `if `test, but as I see how I now have it, it is doing exactly the same thing in terms of what is written to `message`.
````
  if (len + 1 == sizeof(message))
    /* Indicate things are truncated */
    snprintf(&message[len-5], sizeof(message)-len+5, " ...\n");
  else
    snprintf(&message[len], sizeof(message)-len, "\n");
````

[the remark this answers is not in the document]
Comment (boaks) [reply dtls_debug.c:432]: My variant obsoletes the strlen.

`len` is only used to check for overflow, and that is always `len + 1 < sizeof(message)`.
If `len` exceeds message buffer, then `len` is not used but instead the `message buffer size`.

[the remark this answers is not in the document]
Comment (mrdeep1) [reply dtls_debug.c]: OK - gone with your version and pushed the code.  However, it should be &message[sizeof(message)-6] to allow for the trailing zero byte.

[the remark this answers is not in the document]
Comment (boaks) [reply dtls_debug.c]: Yes, I overseen the heading space.

**Comment by boaks** [comment]
@obgm 

Final check from your side and then merging it?

**Comment by boaks** [comment]
Does the build check fail to detect, that "\n" is 1 char not two?

**Comment by boaks** [comment]
> /home/runner/work/tinydtls/tinydtls/dtls_debug.c: In function ‘dtls_dsrv_hexdump_log’:
> /home/runner/work/tinydtls/tinydtls/dtls_debug.c:434:53: error: ‘snprintf’ output truncated before the last format character [-Werror=format-truncation=]
>   434 |     snprintf(&message[sizeof(message)-7], 5, " ...\n");
>       |                                                     ^
> /home/runner/work/tinydtls/tinydtls/dtls_debug.c:434:5: note: ‘snprintf’ output 6 bytes into a destination of size 5
>   434 |     snprintf(&message[sizeof(message)-7], 5, " ...\n");
>       |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> cc1: all warnings being treated as errors
> make[2]: *** [CMakeFiles/tinydtls.dir/build.make:188: CMakeFiles/tinydtls.dir/dtls_debug.c.o] Error 1

my guess: " ...\n" doesn't convert the "\n" and then warns about that.

```
// suppress false clang warning ...
snprintf(&message[sizeof(message)-7], 6, " ...\n");
```

mrdeep1:
I am offline - back at my desk in a couple of hours time - it should be 6 in both places.

boaks:
> I am offline - back at my desk in a couple of hours time - it should be 6 in both places.

Yes, but then I would go for a comment, that the second 6 will be a 5 and 6 is only used because of the warning. Otherwise this will make trouble, if someone changes it, maybe even only in proprietary repos.

mrdeep1:
Yes, it needs to be 6 in both places (to include terminating 0 byte).  Comment added.  Code pushed.

mrdeep1:
Fixed similar overflow issue in `dsrv_log()` and code pushed.

boaks:
I tested 8db0ce0dc39e6d8390b1b5bb87ce3cc790b83a79 successfully with libcoap using the interoperability test of eclipse/californium. So again from my side

LGTM

PR #159, opened 2022-07-25
dtls.c: introduce dtls_writev()

benpicco:
`dtls.c` already provides a `dtls_send_multi()` function to send data from discontinuous buffers.

Also provide this functionality to the user, so e.g. protocol header and payload data can be supplied in separate buffers, removing the requirement for an additional work buffer on the application side.

### DIFF
--- a/dtls.h
+++ b/dtls.h
@@ -291,6 +291,23 @@ int dtls_close(dtls_context_t *ctx, const session_t *remote);
  */
 int dtls_renegotiate(dtls_context_t *ctx, const session_t *dst);
 
+/**
+ * Writes the application data given in multiple buffers to the peer
+ * specified by @p session.
+ *
+ * @param ctx      The DTLS context to use.
+ * @param session  The remote transport address and local interface.
+ * @param buf_array     Array of buffers with the data to write.
+ * @param buf_len_array The length of the arrays in @p buf_array.
+ * @param buf_array_len The number of data arrays.
+ *
+ * @return The number of bytes written, @c -1 on error or @c 0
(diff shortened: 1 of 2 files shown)


boaks:
I guess, you have also an application, which benefits from that.

LGTM.

obgm:
Thanks for this enhancement. Two comments:

1. For backwards compatibility, I suggest to keep `dtls_write()` and just map it to the new function.
2. Naming: Do we want to expose the `_multi` to the public API or go with the traditional `_writev`?

dtls.h
```diff
@@ -303,8 +320,11 @@ int dtls_renegotiate(dtls_context_t *ctx, const session_t *dst);
  * @return The number of bytes written, @c -1 on error or @c 0
  *         if the peer already exists but is not connected yet.
  */
-int dtls_write(struct dtls_context_t *ctx, session_t *session, 
-	       uint8 *buf, size_t len);
+static inline
+int dtls_write(struct dtls_context_t *ctx, session_t *session,
+	       uint8 *buf, size_t len) {
+  return dtls_write_multi(ctx, session, &buf, &len, 1);
+}
```
boaks, line 327: @obgm 

I assumed, this is the function for backwards compatibility.
Does it work for your?

benpicco:
> Do we want to expose the `_multi` to the public API or go with the traditional `_writev`?

+1 for `_writev`, should it also make use of `struct iovec` for good measure?

dtls.h
obgm: Yes, sorry. I got confused by the diff output.

obgm (approved):

PR #160, opened 2022-08-01
dtls.c: consider max identity length in `DTLS_CKXEC_LENGTH`

JKRhb:
When long identities are used in PSK mode, a client is currently not able to handle them during the key exchange even when `DTLS_PSK_MAX_CLIENT_IDENTITY_LEN` is increased since the `DTLS_CKXEC_LENGTH` is not adjusted accordingly.

This PR introduces a potential fix for this problem, using the maximum of either `DTLS_EC_KEY_SIZE * 2` or `DTLS_PSK_MAX_CLIENT_IDENTITY_LEN` for the calculation of `DTLS_CKXEC_LENGTH`.


boaks:
LGTM

obgm (approved):

obgm:
Nice catch!

PR #162, opened 2022-08-01
dtls: redefine `dtls_write` without `static inline`

JKRhb:
While https://github.com/eclipse/tinydtls/pull/159 made a very useful addition by introducing `dtls_writev`, it made `dtls_write` unusable in shared libraries by turning it into a `static inline` function. This PR proposes reintroducing `dtls_write` as a "regular" function when not building for an RTOS (Contiki, RIOT, or Zephyr), making it available in shared libraries again.

Using a doxygen member group, it should be possible to reuse the documentation for both function variants, reducing redundancy.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -331,6 +331,12 @@ dtls_writev(struct dtls_context_t *ctx,
   }
 }
 
+int
+dtls_write(struct dtls_context_t *ctx, session_t *session,
+	       uint8 *buf, size_t len) {
+  return dtls_writev(ctx, session, &buf, &len, 1);
+}
+
 static int
 dtls_get_cookie(uint8 *msg, size_t msglen, uint8 **cookie) {
   /* To access the cookie, we have to determine the session id's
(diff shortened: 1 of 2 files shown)


boaks:
FMPOV, if a static inline function is not working, I would just replace it by a non-static variant without #ifdef .

(My C summit was 30 years ago, so I don't know, if "static inline" works for shared libraries as well and just requires some compiler switches.)

mrdeep1:
I think this is more to do with backwards binary  API comparability.  You have a compiled application using a .so shared library.  Library is updated, but application is not re-complied.  Application fails because dtls_write() is no longer there.

boaks:
My 2 cents.

About this function:

- if we go back, then I would prefer without `ifdef`. 

About backwards binary API compatibility in general:

- this is 0.8.6 ... not 1.5.0
- in my experience, the very most are using it for embedded as static link library
- "backwards binary API compatibility" requires in my opinion also stable data structs in the API. Especially introducing new functionality e.g. selectable cipher suites (CCM/CCM8) in my opinion naturally goes into `dtls_handler_t`,  and will then require to keep the old API as it is and additional add new ones. Not that this is not impossible, but FMPOV, I prefer to spend my time in other topics, which have a better pay off.

boaks:
LGTM

obgm (approved):

PR #163, opened 2022-08-05
github: Add in build and test workflow

mrdeep1:
Unfortunately out of path builds do not work at present.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,64 @@
+name: Build Tests
+
+on:
+  push:
+    branches:
+      - main
+      - develop
+      - gh-workflows
+  pull_request:
+    branches:
+      - main
+      - develop
+
+env:
+  PLATFORM: posix
+  TESTS: yes
+
+jobs:
+  build-linux-autotools:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        CC: ["gcc", "clang" ]
+    steps:
+      - uses: actions/checkout@v3
+      - name: setup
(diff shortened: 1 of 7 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,44 @@
+name: Build Tests
+
+on:
+  push:
+    branches:
+      - main
+      - develop
+      - release-*
```
obgm, line 8: @boaks has suggested not to use the release-branches for tinydtls. As there are currently no releases and no release branches, I suggest removing this entry to avoid confusion or erroneous assumptions.

obgm (commented):
One small comment, otherwise this looks good and works.

.github/workflows/main.yml
mrdeep1: Removed and code pushed with cmake.  clang to follow.

obgm: Thanks. I would expect that clang can just be added to the compile matrix (line 23).

mrdeep1:
Hmm, just added in clang to the matrix, pushed the code and now the tests are not running.

mrdeep1:
My bad, fixed a bad change, tests are now working.

obgm (approved):
LGTM

boaks:
LGTM

mrdeep1:
By default 90 days.  github repository -> Settings -> Actions -> General ->Artifact and log retention

boaks:
Thanks, lets see, if 90 days is a proper value ;-).

PR #165, opened 2022-09-20
Fix cmake warnings of zephyr builds.

boaks:
Exclude cmake options for zephyr builds, if these are set by zephyr.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,7 +25,12 @@ project(tinydtls)
 
 include (AutoConf.cmake)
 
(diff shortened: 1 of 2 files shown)


boaks:
@obgm 

If you could have a short look. 
it affects only the zephyr build, replacing the cmake options by variables set from kconfig.
The fixes some cmake warnings.

boaks:
I turns out, that the `file(CONFIGURE ...)` requires a newer cmake version.
This PR therefore only applies this, if cmake has the required version.

PR #167, opened 2022-09-26
LwIP: Support LwIP builds when NO_SYS = 1

mrdeep1:
Follow the model of how things are done for Contiki.

Set up a new WITH_LWIP_NO_SOCKET (which happens with NO_SYS = 1)
and use this when setting up addresses.

Add in platform specific for dtls_prng_lwip.c and lwip_platform.h

Update dtls_time.[ch] to handle LWIP version of time.

Code works for NO_SYS = 1 or NO_SYS = 0.

### DIFF
--- a/dtls_debug.c
+++ b/dtls_debug.c
@@ -182,7 +182,7 @@ dsrv_print_addr(const session_t *addr, char *buf, size_t len) {
   len -= err;
 
   return p - buf;
-#else /* HAVE_INET_NTOP */
+#else /* ! HAVE_INET_NTOP */
 
 #ifdef WITH_CONTIKI
   char *p = buf;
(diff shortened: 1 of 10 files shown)


dtls_debug.c
```diff
@@ -130,7 +130,7 @@ dtls_strnlen(const char *s, size_t maxlen) {
  */
 static size_t
 dsrv_print_addr(const session_t *addr, char *buf, size_t len) {
-#ifdef HAVE_INET_NTOP
+#if defined(HAVE_INET_NTOP) && !defined(WITH_LWIP)
```
boaks, line 133: Just as general remark:
if the "auto-detection" of cmake/autoconf fails for some platforms, maybe placing the "os fixes" in "tinydtls.h" helps to remove that stuff from other code. (Not with this PR, but may be afterward.) 

What do others think about that?

mrdeep1: I did struggle with this one. 

My non-LwIP specific build environment did find inet_ntop and hence HAVE_INET_NTOP.  Looking to see if I can easily make the code cleaner.

boaks: OK, so it's more caused by "cross-compiling".

boaks:
I have no setup to test it, but the code itself looks good.

So LGTM from my side.

boaks:
@obgm 

Any comments/issues from your side?

PR #170, opened 2022-09-28
treewide: make type casts explicit

JKRhb:
As another follow-up to #126, when compiling with Visual Studio, there are a lot of warnings regarding a potential data loss due to implicit casts with a potential lossy conversion.

This PR makes the type casts in question explicit, getting rid of potential warnings.

### DIFF
--- a/tests/unit-tests/test_ecc.c
+++ b/tests/unit-tests/test_ecc.c
@@ -129,7 +129,7 @@ t_test_ecc_dh(void) {
   uint32_t tempBy2[8];
   uint32_t secretA[8];
   uint32_t secretB[8];
-  int ret;
+  size_t ret;
 
   ret = dtls_prng((void *)secretA, sizeof(secretA));
   CU_ASSERT(ret > 1);


obgm:
Thanks for working on this. The warnings usually are there for a reason, and shutting them up with type casts is not always the best solution. One example is the return type of `dtls_ccm_encrypt_message()` which seems to be wrong (the result of adding two `size_t` values obviously should be `size_t`; The explicit typecast therefore should result in another warning...).
For this (and some other places), I recommend to double-check if there are no cleaner ways to address the compiler warnings.

JKRhb:
Thank you for your feedback, @obgm! I will revisit the casts and try to come up with a cleaner solution :)

JKRhb:
Just a quick update: I now managed to get rid of most of the type casts while keeping stricter compiler settings (adding the `-Wconversion` and `-Wno-sign-conversion` flags). However, this required changing some of the return types from `int` to `ssize_t` in order to be able to cover -1 as a return value. Is using `ssize_t` fine in general (since it is not a C99, but a POSIX type)? Or would a different type (such as `intmax_t`) be preferable here?

PR #171, opened 2022-09-28
doc: Correct Doxygen information

mrdeep1:
Dynamically build DoxygenLayout.xml to get the latest layout for the current
version of doxygen.

Update Doxyfile.in to the Ubuntu20 version.


obgm:
> latest correct layout Update Doxyfile.in 

Hmmm:

```
$ doxygen -v
1.9.4
```

obgm:
Thinking about it a little more, since the build system is Ubuntu 20, it makes sense to use the doxygen version that comes with that release. I just would not call it "latest" :-)

mrdeep1:
Hopefully corrected the commit message to reflect the current situation.

crypto.h
```diff
@@ -205,11 +205,15 @@ typedef struct {
...
+ * \param label    The label.
+ * \param labellen Length of \p label.
+ * \param random1 The random pt 1. 
+ * \param random1len Length of \p random1.
+ * \param random2 The random pt 2. 
+ * \param random2len Length of \p random2.
  * \param buf     Output buffer where the result is XORed into
- *                The buffe must be capable to hold at least
- *                \p buflen bytes.
+ * \param buflen  THe available space for \p buf
```
obgm, line 215: "THe" → "The"

obgm (commented):
One small typo left.

crypto.h
mrdeep1: Corrected and change pushed.

obgm:
LGTM

PR #179, opened 2022-11-24
global.h: enable constrained stack if DTLS_MAX_BUF is too large

benpicco:
When bumping `DTLS_MAX_BUF` to encrypt larger messages (e.g. to be sent over Ethernet) the size of the stack-allocated `sendbuf` will also grow.

This can lead to surprising crashes / stack corruption as thread stacks are usually not that large.

To prevent this, automatically enable `DTLS_CONSTRAINED_STACK` if `DTLS_MAX_BUF` is beyond a given threshold (default size with `DTLS_ECC`).

### DIFF
--- a/global.h
+++ b/global.h
@@ -57,6 +57,15 @@ typedef unsigned char uint48[6];
 #endif /* WITH_CONTIKI || RIOT_VERSION */
 #endif
 
+/*
+ * DTLS send buf is alloctaed on the stack by default
+ */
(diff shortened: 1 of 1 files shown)


boaks:
LGTM

obgm (approved):
This type of automated configuration changes is a bit dangerous, especially without user notification. Since constrained devices are generally expected to use a constrained stack anyway, I think this is okay.

Issue #175, opened 2022-10-31
OpenSSL 3.1 fails to connect to TinyDTLS server


mrdeep1:
With the server running as
````
tests/dtls-server -v999 -p 4433 -A 127.0.0.1
````

This works fine for OpenSSL < 3.0, but fails for >= 3.0

````
openssl s_client -psk 73656372657450534b -psk_identity Client_identity -dtls -cipher 'PSK:!NULL' -connect 127.0.0.1:4433
````
It appears that support for RFC 5746 is needed as per [CAN-2009-3555](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CAN-2009-3555).

boaks:
Openssl is anyway a bad choice for IoT ;-).

mrdeep1:
That maybe, but my (PSK) regression tests are all messed up when running Ubuntu 22!

boaks:
I'm running still 20.04. 
If possible, please provide a capture. Maybe the tinydtls server doesn't proper ignore it.
Or do you test a "renegotiation" for DTLS?

boaks:
`SSL_OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION` maybe a the fast way with openssl 3.0.

Not sure, how to proceed with that in general:

In my opinion (the discussion have been a couple of years ago in Eclipse/Californium), renegotiation was in too many cases misinterpreted as a way to overcome broken DTLS in the context of NATs. It doesn't work there.
With UDP, we never need a TCP connect and with that, it doesn't make a difference, if the client uses a fresh dtls-handshake or one encrypted by the current DTLS context.

So, the first question is more:
- who is using the "renegotiation" feature and for which purpose?
- depending on the answers, it may be easier (and code saving) to remove the left overs. 
  Or to try to fix it in the sense of RFC5746.
 
My interpretation of RFC5746 is, that it doesn't support to implement "no renegotiation at all". It looks, like the way to do that is to "negotiate the secure renegotiation" and then block, when that is really requested. Depending  on the outcome, if renegotiation must be supported, I will try to ask the tls list about the way, to indicate, that renegotiation isn't supported at all.

mrdeep1:
> If possible, please provide a capture and logs.
> Maybe the tinydtls server doesn't proper ignore it.

[traces.zip](https://github.com/eclipse/tinydtls/files/9902499/traces.zip)

> Or do you test a "renegotiation" for DTLS?

Not that I am aware of.

[...]

PR #181, opened 2022-11-29
Rfc5746

boaks:
Issue #175

### DIFF
--- a/crypto.h
+++ b/crypto.h
@@ -140,6 +140,7 @@ typedef struct dtls_user_parameters_t {
    */
   dtls_cipher_t cipher_suites[DTLS_MAX_CIPHER_SUITES + 1];
   unsigned int force_extended_master_secret:1; /** force extended master secret extension (RFC7627) */
+  unsigned int force_renegotiation_info:1;     /** force renegotiation info extension (RFC5746) */
 } dtls_user_parameters_t;
 
 typedef struct {
@@ -159,6 +160,7 @@ typedef struct {
   dtls_cipher_index_t cipher_index;		/**< internal index for cipher_suite_params, DTLS_CIPHER_INDEX_NULL for TLS_NULL_WITH_NULL_NULL */
   unsigned int do_client_auth:1;
   unsigned int extended_master_secret:1;
+  unsigned int renegotiation_info:1;
   union {
 #ifdef DTLS_ECC
     dtls_handshake_parameters_ecdsa_t ecdsa;
--- a/dtls.c
+++ b/dtls.c
@@ -140,6 +140,9 @@ memarray_t dtlscontext_storage;
  * ec point format        := 6 bytes   => 26
  * sign. and hash algos   := 8 bytes
  * extended master secret := 4 bytes   => 12
+ *
+ * (The ClientHello uses TLS_EMPTY_RENEGOTIATION_INFO_SCSV
+ *  instead of renegotiation info)
  */
 #define DTLS_CH_LENGTH sizeof(dtls_client_hello_t) /* no variable length fields! */
 #define DTLS_COOKIE_LENGTH_MAX 32
@@ -660,6 +663,7 @@ static const dtls_user_parameters_t default_user_parameters = {
     },
 #endif /* DTLS_DEFAULT_CIPHER_SUITES */
   .force_extended_master_secret = 1,
+  .force_renegotiation_info = 1,
 };
 
 /** only one compression method is currently defined */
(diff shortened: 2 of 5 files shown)


boaks:
The client always include the TLS_EMPTY_RENEGOTIATION_INFO_SCSV and the server answers with an empty renegotiation info, if either the TLS_EMPTY_RENEGOTIATION_INFO_SCSV or a empty renegotiation info is contained in the Client_Hello. 
That enables to execute handshakes with implementations, which requires RFC5746 support.

For now, it's optional in tinydtls, if the other peer supports it. 

I will create a second PR, which adapts the `get_cipher_suites` into  a `get_dtls_paramater`and a specific `struct dtls_user_parameter` and add a flags "renegotiation_info_required" there.

boaks:
@mrdeep1 

If you find the time, you may test it against OpenSsl 3.1.

mrdeep1:
Will do - it will be later today

mrdeep1:
Unfortunately, it fails to match a cipher suite and so gives up.  Using original `c84e36f `build for TinyDTLS, all works.  It makes no difference if OpenSSL (with or without the fix in place) or MbedTLS is used as a client for the test.  However GnuTLS does work - hmm.

[test.zip](https://github.com/eclipse/tinydtls/files/10113912/test.zip) has a pcap and log file of the failure.

mrdeep1:
With this fix, things progress, but a bad server_hello (not server_hello_done) is reported by the client which then raises an alert.
````
diff --git a/dtls.c b/dtls.c
index 8e8dc13..9bde81a 100644
--- a/dtls.c
+++ b/dtls.c
@@ -1259,7 +1259,7 @@ dtls_update_parameters(dtls_context_t *ctx,
       config->renegotiation_info = 1;
     } else {
       config->cipher_index = get_cipher_index(config->cipher_suites, dtls_uint16_to_int(data));
-      ok = known_cipher(ctx, config->cipher_index, 0);
+      ok |= known_cipher(ctx, config->cipher_index, 0);
     }
     i -= sizeof(uint16);
     data += sizeof(uint16);
````
See logs 
[test1.txt](https://github.com/eclipse/tinydtls/files/10114520/test1.txt)

**Comment by mrdeep1** [comment]
Seems to be working with this:-
````
diff --git a/dtls.c b/dtls.c
index 8e8dc13..ef98ecb 100644
--- a/dtls.c
+++ b/dtls.c
@@ -1257,7 +1257,7 @@ dtls_update_parameters(dtls_context_t *ctx,
   while ((i >= (int)sizeof(uint16)) && (!ok || !config->renegotiation_info)) {
     if (dtls_uint16_to_int(data) == TLS_EMPTY_RENEGOTIATION_INFO_SCSV) {
       config->renegotiation_info = 1;
-    } else {
+    } else if (!ok) {
       config->cipher_index = get_cipher_index(config->cipher_suites, dtls_uint16_to_int(data));
       ok = known_cipher(ctx, config->cipher_index, 0);
     }
````

File: crypto.c [hunk 1]
```diff
@@ -157,12 +157,14 @@ dtls_handshake_parameters_t *dtls_handshake_new(void)
 
   memset(handshake, 0, sizeof(*handshake));
 
-  /* initialize the handshake hash wrt. the hard-coded DTLS version */
-  dtls_debug("DTLSv12: initialize HASH_SHA256\n");
-  /* TLS 1.2:  PRF(secret, label, seed) = P_<hash>(secret, label + seed) */
-  /* FIXME: we use the default SHA256 here, might need to support other
-            hash functions as well */
-  dtls_hash_init(&handshake->hs_state.hs_hash);
+  if (handshake) {
```
mrdeep1, line 160: No need for this test as handshake cannot be NULL at this point (line 153).

mrdeep1:
With [84f834a](https://github.com/eclipse/tinydtls/commit/84f834a69eefefd4180edcda5a1b485cadb75b07) libcoap regression tests work for when `SSL_CTX_set_options(context->dtls.ctx, SSL_OP_LEGACY_SERVER_CONNECT )` is set or not for OpenSSL.  So, this LGTM.

boaks:
Thanks for testing!

The current plan is to have a final review and merge the PR chain "step by step", starting with PR #147 and #148. Depending on the time needed for #147, we may merged that first to "develop".

obgm (approved):
LGTM.

There is a minor detail that might lead to a buffer overrun in dtls_send_client_hello(). I will open a PR for that right after merging this.

PR #182, opened 2022-12-02
User parameter

boaks:


dtls.c
```diff
@@ -588,26 +588,28 @@ dtls_set_handshake_header(uint8 type,
...
-    TLS_PSK_WITH_AES_128_CCM,
+      TLS_PSK_WITH_AES_128_CCM_8,
+      TLS_PSK_WITH_AES_128_CCM,
 #endif /* DTLS_PSK */
-    /* RFC 5746, pseudo cipher suite to indicate secure renegotiation */
-    TLS_EMPTY_RENEGOTIATION_INFO_SCSV,
     /* TLS_NULL_WITH_NULL_NULL must always be the last entry as it
      * indicates the stop marker for the traversal of this table. */
-    TLS_NULL_WITH_NULL_NULL
+      TLS_NULL_WITH_NULL_NULL
```
obgm, line 607: Now the comment above this line is not aligned anymore.

tests/dtls-client.c
```diff
@@ -216,13 +216,26 @@ send_to_peer(struct dtls_context_t *ctx,
...
+get_user_parameters(struct dtls_context_t *ctx,
+    session_t *session, dtls_user_parameters_t *user_parameters) {
   (void) ctx;
   (void) session;
-  *cipher_suites = ciphers;
+  user_parameters->force_extended_master_secret = force_extended_master_secret;
+  user_parameters->force_renegotiation_info = force_renegotiation_info;
+  if (ciphers) {
+    int index = 0;
+    while (index <= DTLS_MAX_CIPHER_SUITES) {
```
obgm, line 231: I think that this loop only does the correct thing if ciphers does not contain more than DTLS_MAX_CIPHER_SUITES elements because otherwise, the terminating TLS_NULL_WITH_NULL_NULL will be copied.  Is this guaranteed?

tests/dtls-server.c
```diff
@@ -163,13 +169,26 @@ send_to_peer(struct dtls_context_t *ctx,
...
+get_user_parameters(struct dtls_context_t *ctx,
+    session_t *session, dtls_user_parameters_t *user_parameters) {
   (void) ctx;
   (void) session;
-  *cipher_suites = ciphers;
+  user_parameters->force_extended_master_secret = force_extended_master_secret;
+  user_parameters->force_renegotiation_info = force_renegotiation_info;
+  if (ciphers) {
+    int index = 0;
+    while (index <= DTLS_MAX_CIPHER_SUITES) {
```
obgm, line 184: same as in dtls-client.c

obgm:
Thank you very much for this proposal. I think that combining the user parameters into a single callback function results in a more compact and comprehensible configuration.

tests/dtls-client.c
boaks: The condition protects the destination array `dtls_cipher_t cipher_suites[DTLS_MAX_CIPHER_SUITES + 1];` from overflow.

But I guess, it must be corrected to `index < DTLS_MAX_CIPHER_SUITES` to ensure, that the last entry is the already stored TLS_NULL_WITH_NULL_NULL.

boaks:
Now included in PR #148 and PR #181 .

PR #184, opened 2022-12-07
Fix tests apps

boaks:


tests/dtls-client.c
```diff
@@ -147,8 +161,8 @@ get_psk_info(struct dtls_context_t *ctx UNUSED_PARAM,
 #ifdef DTLS_ECC
 static int
 get_ecdsa_key(struct dtls_context_t *ctx,
-	      const session_t *session,
```
obgm, line 150: This is the correct indentation. Why would you change that? (Same for the following couple of functions.)

boaks: The old format used a "tab" and replacing that with 2 spaces is my interpretation of "For better reading the indentation is set to 2 characters as spaces".

Maybe we better try to precise, how a "multi-line  function definition  should look like. For the declaration it seems to align with the name, but that doesn't work and may even be a misinterpretation of the example.

mrdeep1: I concur that in general TABs should be replaced by 2 spaces, but then indentation needs to be considered.  As in the second line of
````
  return sendto(fd, data, len, MSG_DONTWAIT,
		&session->addr.sa, session->size);
````
I would expect the indentation of the second and subsequent line indentations to be similar for function definitions, namely aligned to be under the character following the opening `(` as per
````
static int
get_ecdsa_key(struct dtls_context_t *ctx,
	      const session_t *session,
````

boaks: @obgm 

Is that proposal OK for you?
If so, I will change the spaces here and amend that guidance to the contribution guide.

obgm: > Is that proposal OK for you? If so, I will change the spaces here and amend that guidance to the contribution guide.

Yes, very good. Sorry, somehow, I missed that there have been Tab characters in the code. Looks as if we all are on the same page here, so please, go ahead: LGTM.

PR #185, opened 2023-01-24
Amend source formatting with rule for function definitions with multiple lines

boaks:
See discussion in PR #184.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -116,16 +116,25 @@ dtls_get_peer(const dtls_context_t *ctx, const session_t *session) {
 ...
 }
 ```
+* If in this case the line with the function name and parameter-list exceeds
+  the line length of 80 characters caused by that parameter-list, then split
+  this list on multiple lines aligning with the first parameter.
 
+```
+int
+dtls_read(const dtls_context_t *ctx, const session_t *session,
+          uint8_t* buffer, size_t buffer_length) {
+...
+}
+```
 * Declarations in header files do not follow the previous rule. For
   example, the declaration for `dtls_get_peer()` in `dtls.h` reads as
   follows:
 
 ```
 dtls_peer_t *dtls_get_peer(const dtls_context_t *context,
-			   const session_t *session);
+             const session_t *session);
 ```
-
 * A useful source code documentation is mandatory. Mostly to be done
   within the source code files.
 


CONTRIBUTING.md
```diff
@@ -116,14 +116,24 @@ dtls_get_peer(const dtls_context_t *ctx, const session_t *session) {
...
+}
+```
 * Declarations in header files do not follow the previous rule. For
   example, the declaration for `dtls_get_peer()` in `dtls.h` reads as
   follows:
 
 ```
 dtls_peer_t *dtls_get_peer(const dtls_context_t *context,
-			   const session_t *session);
+             const session_t *session);
```
mrdeep1, line 136: This is not right.  Previous TABS have been converted to double spaces, messing up the indentation of the second line.

CONTRIBUTING.md
```diff
@@ -116,14 +116,24 @@ dtls_get_peer(const dtls_context_t *ctx, const session_t *session) {
...
+  first parameter.
 
+```
+int
+dtls_read(const dtls_context_t *ctx, const session_t *session,
+          uint8_t* buffer, size_t buffer_length) {
+...
+}
+```
 * Declarations in header files do not follow the previous rule. For
```
mrdeep1, line 130: This has now become ambiguous.  It refers to the return type being on the preceding line, not how you split long lines.

* Declarations in header files keep the return type on the same line as the function definition.  For

boaks: Just to mention:
I'm not sure, which format should be applied for function declarations in *.c files.

mrdeep1: As these functions all should be prefixed by `static`, this increases the line length requirement.  My suggestion for the .c files is that the static + return type are kept on the previous line.

CONTRIBUTING.md
mrdeep1: I would have expected the first character of the second line (and subsequent lines if appropriate) to be aligned with the first parameter of the parameter list in the same way as it is done in the *.c files.

One of the things we have been looking at is [clang-format](https://clang.llvm.org/docs/ClangFormat.html) to be consistent with styles.  clang-format is pretty flexible in its style configurations, so lots of room for negotiation there!

obgm (approved):
LGTM

(In good old times, line lengths used to be less than 80 characters to leave some room for prefixing in emails or when passing patches around. Nowadays, people seem to care less about this.)

PR #187, opened 2023-02-05
RIOT: make use od module for hexdump

benpicco:
While trying to figure out why tinydtls still requires `gettimeofday()` even though `dtls_ticks()` uses ZTimer now, I found that it's still used via `time()` in `dtls_dsrv_hexdump_log()`.

Replace it with a native hexdump function as it was done for Zephyr.

This allows us to drop the `gettimeofday` compat modules that was so far needed to build tinydtls with RIOT.

#### before

``` 
   text	   data	    bss	    dec	    hex	filename
 123068	    384	  24920	 148372	  24394	/home/benpicco/dev/RIOT/examples/dtls-sock/bin/openmote-b/dtls_sock.elf
```

#### with this patch

```
   text	   data	    bss	    dec	    hex	filename
 107756	    272	  24776	 132804	  206c4	/home/benpicco/dev/RIOT/examples/dtls-sock/bin/openmote-b/dtls_sock.elf
```

### DIFF
--- a/dtls_debug.h
+++ b/dtls_debug.h
@@ -30,6 +30,7 @@
 
 #ifdef RIOT_VERSION
 #include "log.h"
+#include "od.h"
 #endif
 
 #ifdef WITH_CONTIKI
@@ -138,8 +139,8 @@ void dtls_dsrv_log_addr(log_t level, const char *name, const session_t *addr);
 #define dtls_notice(...) LOG_INFO(__VA_ARGS__)
 #define dtls_info(...) LOG_INFO(__VA_ARGS__)
 #define dtls_debug(...) LOG_DEBUG(__VA_ARGS__)
-#define dtls_debug_hexdump(name, buf, length) dtls_dsrv_hexdump_log(DTLS_LOG_DEBUG, name, buf, length, 1)
-#define dtls_debug_dump(name, buf, length) dtls_dsrv_hexdump_log(DTLS_LOG_DEBUG, name, buf, length, 0)
+#define dtls_debug_hexdump(name, buf, length) { if (LOG_DEBUG <= LOG_LEVEL) { LOG_DEBUG("-= %s (%zu bytes) =-\n", name, length); od_hex_dump(buf, length, 0); }}
+#define dtls_debug_dump(name, buf, length) { if (LOG_DEBUG <= LOG_LEVEL) { LOG_DEBUG("%s (%zu bytes):", name, length); od_hex_dump(buf, length, 0); }}
 #else /* neither RIOT nor Zephyr */
 #define dtls_emerg(...) dsrv_log(DTLS_LOG_EMERG, __VA_ARGS__)
 #define dtls_alert(...) dsrv_log(DTLS_LOG_ALERT, __VA_ARGS__)


dtls_debug.h
```diff
@@ -138,8 +139,8 @@ void dtls_dsrv_log_addr(log_t level, const char *name, const session_t *addr);
 #define dtls_notice(...) LOG_INFO(__VA_ARGS__)
 #define dtls_info(...) LOG_INFO(__VA_ARGS__)
 #define dtls_debug(...) LOG_DEBUG(__VA_ARGS__)
-#define dtls_debug_hexdump(name, buf, length) dtls_dsrv_hexdump_log(DTLS_LOG_DEBUG, name, buf, length, 1)
-#define dtls_debug_dump(name, buf, length) dtls_dsrv_hexdump_log(DTLS_LOG_DEBUG, name, buf, length, 0)
+#define dtls_debug_hexdump(name, buf, length) { if (LOG_DEBUG <= LOG_LEVEL) { LOG_DEBUG("%s (%zu bytes):", name, length); od_hex_dump(buf, length, 0); }}
+#define dtls_debug_dump(name, buf, length) { if (LOG_DEBUG <= LOG_LEVEL) { LOG_DEBUG("%s (%zu bytes):", name, length); od_hex_dump(buf, length, 0); }}
```
boaks, line 143: For tinydtls the difference of `dtls_debug_hexdump` and `dtls_debug_dump` is, that the first logs multiple lines (suitable for larger messages), while the second logs a single line. 
In zephyr I didn't found something to achieve that behavior.
For RIOT it may be possible, using "\n" at the end of the LOG_DEBUG and then use a proper `width` for each instead of the 0. If you don't feel well with that proposal, the current version will also do it for me.

benpicco: `od_hex_dump()` will already split the buffer into multiple lines on it's own. A width of 0 will just select the default with.

I added a newline to `dtls_debug_hexdump()`.

boaks:
Just to mention: don't care about the ECA error. The eclipse admin are already on it. It's a general issue.

benpicco:
Looks like it's resolved :)

boaks:
Yes, from my side LGTM.

dtls_debug.h
```diff
@@ -128,7 +129,7 @@ void dtls_dsrv_log_addr(log_t level, const char *name, const session_t *addr);
 #define dtls_notice(...) LOG_INF(__VA_ARGS__)
 #define dtls_info(...) LOG_INF(__VA_ARGS__)
 #define dtls_debug(...) LOG_DBG(__VA_ARGS__)
-#define dtls_debug_hexdump(name, buf, length) { LOG_DBG("%s (%zu bytes):", name, length); LOG_HEXDUMP_DBG(buf, length, name); }
+#define dtls_debug_hexdump(name, buf, length) { LOG_DBG("-= %s (%zu bytes) =-\n", name, length); LOG_HEXDUMP_DBG(buf, length, name); }
```
obgm, line 132: I do not understand what this change is for and why it is in a RIOT-related PR?

obgm: @boaks As the one managing and using the Zephyr support, please decide on this one.

LGTM.

boaks: I guess, that is accidentally change here. According the comment below, it should be in the RIOT part.

@benpicco 
Could you please move it to the RIOT lines below?

benpicco: Oops, good catch - I accidentally added the newline to the wrong line

PR #188, opened 2023-02-06
dtls_prng_posix.c: fail for posix without HAVE_GETRANDOM.

boaks:
Add #error with explanation and hint to solve it.

### DIFF
--- a/platform-specific/dtls_prng_posix.c
+++ b/platform-specific/dtls_prng_posix.c
@@ -37,7 +37,31 @@ int
 dtls_prng(unsigned char *buf, size_t len) {
 #ifdef HAVE_GETRANDOM
   return getrandom(buf, len, 0);
-#else /* !HAVE_GETRANDOM */
+#elif defined(HAVE_RANDOM)
+
+#define RAND_BYTES (RAND_MAX >= 0xffffff ? 3 : (RAND_MAX >= 0xffff ? 2 : 1))
+
+  if (len) {
+    size_t klen = len;
+    uint8_t byte_counter = RAND_BYTES;
+    uint32_t rand = random();
+    while (1) {
+      *buf++ = rand & 0xFF;
+      if (!--klen) {
+        break;
+      }
+      if (--byte_counter) {
+        rand >>= 8;
+      } else {
+        rand = random();
+        byte_counter = RAND_BYTES;
+      }
+    }
+  }
+  return len;
+#else /*!HAVE_GETRANDOM && !HAVE_RANDOM */
+  #error "CVE-2021-34430: using rand() for crypto randoms is not secure!"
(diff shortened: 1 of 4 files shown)


boaks:
@mrdeep1

The failed build for macOS indicates, that "getrandom()" is missing.
Do you know, what is provided there?

mrdeep1:
I think you need to use `random()`.

boaks:
Yes, random() seems to work for macOS.

platform-specific/dtls_prng_posix.c
```diff
@@ -37,7 +37,14 @@ int
 dtls_prng(unsigned char *buf, size_t len) {
 #ifdef HAVE_GETRANDOM
   return getrandom(buf, len, 0);
-#else /* !HAVE_GETRANDOM */
+#elif defined(HAVE_RANDOM)
+  size_t klen = len;
+  while (len--)
+    *buf++ = random() & 0xFF;
+  return klen;
```
mrdeep1, line 44: random() gives 31 bits of randomness.  To make this more efficient, I would suggest
````
+  size_t klen = len;
+  u_int16_t r_value = (uint16_t)random();
+  
+  while (len > sizeof(r_value)) {
+    memcpy(dst, &r_value, sizeof(r_value));
+    len -= sizeof(r_value);
+    dst += sizeof(r_value);
+    r_value = (uint16_t)random();
+  }
+  memcpy(dst, &r_value, len);
+  return klen;
````
Otherwise this PR looks good to me.

Comment (boaks) [reply platform-specific/dtls_prng_posix.c:44]: According [19.8.1 ISO C Random Number Functions](https://www.gnu.org/software/libc/manual/html_node/ISO-Random.html), the value range is`0-RAND_MAX` with `RAND_MAX` from 2147483647 to 32767. So, I'm not sure, if that pays off.

File: platform-specific/dtls_prng_posix.c [hunk 2]
```diff
@@ -37,7 +37,31 @@ int
...
+  if (len) {
+    size_t klen = len;
+    uint8_t byte_counter = RAND_BYTES;
+    uint32_t rand = random();
+    while (1) {
+      *buf++ = rand & 0xFF;
+      if (!--klen) {
+        break;
+      }
+      if (byte_counter--) {
```
mrdeep1, line 53: I think the following is better and more readable (as RAND_BYTES is what it says on the tin)
````
#define RAND_BYTES (RAND_MAX >= 0xffffff ? 3 : (RAND_MAX >= 0xffff ? 2 : 1))

....

      if (--byte_counter) {
````

boaks: Thanks. Done.

obgm (approved):
LGTM

PR #189, opened 2023-02-27
dtls_debug: mock dsrv_log() when NDEBUG is set

benpicco:
Setting `NDEBUG` already replaces `dtls_dsrv_hexdump_log()`, `dtls_dsrv_log_addr()` etc with dummy implementations.

This extends this to also replace `dsrv_log()` and `dtls_{get,set}_log_level()` with a no-op implementation.

### DIFF
--- a/dtls_debug.c
+++ b/dtls_debug.c
@@ -51,8 +51,6 @@ typedef int in_port_t;
 LOG_MODULE_REGISTER(TINYDTLS, CONFIG_TINYDTLS_LOG_LEVEL);
 #endif /* WITH_ZEPHYR */
 
-static int maxlog = DTLS_LOG_WARN;      /* default maximum log level */
-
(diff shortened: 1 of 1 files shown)


boaks:
:-) how many bytes will be saved?

LGTM.

benpicco:
Actually none on the RIOT integration because it already doesn't use those functions.
The main motivation is to not having to *compile* them as the `dsrv_log()` implementations collides with our `printf()` wrapper that automatically places the format string in FLASH on AVR.
This requires the format string to be a literal which this implementation does not satisfy.

With this (and `NDEBUG`) build work on AVR (see https://github.com/RIOT-OS/RIOT/pull/19346)

obgm:
LGTM

obgm (approved):

PR #190, opened 2023-03-11
dtls.c: fix processing of alerts.

boaks:
If alerts remove the peer, additional records need to be dropped.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -4201,7 +4201,7 @@ dtls_handle_message(dtls_context_t *ctx,
 		    uint8 *msg, int msglen) {
   dtls_peer_t *peer = NULL;
   unsigned int rlen;		/* record length */
-  uint8 *data; 			/* (decrypted) payload */
+  uint8 *data = NULL;		/* (decrypted) payload */
   int data_length;		/* length of decrypted payload
 				   (without MAC and padding) */
   int err;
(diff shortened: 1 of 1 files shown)


dtls.c
```diff
@@ -4267,6 +4267,7 @@ dtls_handle_message(dtls_context_t *ctx,
   }
 
   while ((rlen = is_record(msg,msglen))) {
+
```
obgm, line 4270: Why an empty line? I think this should be removed before merge.

dtls.c
```diff
@@ -4395,8 +4407,8 @@ dtls_handle_message(dtls_context_t *ctx,
       } else {
         dtls_stop_retransmission(ctx, peer);
       }
-      break;
-
+      /* no more valid records after alerts */
+      return 0;
```
obgm, line 4411: There is [one case](https://github.com/eclipse/tinydtls/pull/190/commits/62c8c08850df36da7201871eb1e10b2d7c58268b#diff-947f252e5a6242133f45e2963433822253886eb9a6bf4c12c7163ac8e90fe017L4396) where `peer` is currently not set to `NULL` (which is correct, I think). Changing `break` to `return 0` here should not break things, but I wanted to mention this case in order to do a double-check.

dtls.c
```diff
@@ -4282,6 +4283,17 @@ dtls_handle_message(dtls_context_t *ctx,
                  content_type, epoch, pkt_seq_nr, rlen);
     }
 
+    if (!peer) {
```
obgm, line 4286: Since re-validation is triggered always at the end of this while-loop, we could decide to move the initial retrieval into this loop.

obgm (approved):
LGTM, just a couple of minor comments to take into consideration.

dtls.c
boaks: We can keep the old function for DTLS_ALERT_LEVEL_WARNING. But that would only make a difference, if such a Alert/Warning and additional records are sent within the same UDP message.
I will change it.

dtls.c
boaks: OK.

dtls.c
boaks: I kept the copy for the different logging message. 
That could be also achieved, when the snippets are merged.
I will try it.

boaks:
The comments are addressed.

dtls.c
obgm: Correct, and I think that a message that contains additional records after the alert can be treated as broken—why would an application do that?

obgm:
Still looks good :-)
I think we are ready to merge.

boaks:
> I think we are ready to merge.

Yeap, ones the test builds have succeeded :-).

PR #193, opened 2023-04-19
dtls.c: Reduce logging for unsupported tls extensions

mrdeep1:
Every unsupported tls extension is logged using dtls_warn(), and the default logging level is set to DTLS_LOG_WARN.

Update logging call to dtls_notice(), to reduce logging output noise.


boaks:
LGTM

obgm (approved):
LGTM

PR #195, opened 2023-04-24
.h files: Rename to make dtls specific

mrdeep1:
Remove include clashes, in particular with files named as global.h and session.h when used with other projects.

Future work could be to move the header files into a `include/tinydtls` directory, so that all of the includes are of the form "typydtls/dtls_global.h" etc.


boaks:
While I appreciate improvements, I'm not sure, if changing that many files with a couple of large/mid size PRs open. 
Luckily, this project is overall "tiny", so I don't see any real issue with it.

LGTM.

obgm:
I completely agree with @boaks 
We can fix the conflicts and merge this PR once the big PRs are pulled in.

Issue #196, opened 2023-04-27
Three out-of-bound pointers (check_certificate_request)


bathooman:
### Description
- Type: Out of bound pointer
- Priority: High
- The bug affects the develop branch
- This affects the processing of the messages on the client-side.
---------------------------------------------------------------
**How to trigger** 

![image](https://user-images.githubusercontent.com/3464382/234988011-eeac8f68-52c1-493e-89a0-060786686f7a.png)

Upon receipt of a Certificate Request message, if the message is maliciously crafted in a way that the value of the `Certificate Types Count` and `Signature Hash Algorithms Length` are not within the boundary of the `Fragment Length`, three out-of-bound pointer de-reference occurs in the following three points:

1. https://github.com/eclipse/tinydtls/blob/main/dtls.c#L3529
2. https://github.com/eclipse/tinydtls/blob/main/dtls.c#L3531
3. https://github.com/eclipse/tinydtls/blob/main/dtls.c#L3508

My suggestion:

There should always be a check that ensures that the fields that represent the size of fields are within the boundary of [data_length](https://github.com/eclipse/tinydtls/blob/main/dtls.c#L3467).

boaks:
Thanks a lot for reporting!

I see, this function doesn't obey the already processed bytes by subtracting them from  `data_length`.
And, as you report, the size of the length field must also be obeyed, e.g. in [L3510](https://github.com/eclipse/tinydtls/blob/main/dtls.c#L3510), where `data + 1` is used instead of `data + sizeof(uint16)`.

I will provide a fix.

bathooman:
It appears that this bug is fixed in #197. We can close this issue.

PR #197, opened 2023-04-28
Issue 196

boaks:
Fixes issue #196

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -178,16 +178,39 @@ memarray_t dtlscontext_storage;
 #define HANDSHAKE(M) ((dtls_handshake_header_t *)((M) + DTLS_RH_LENGTH))
 #define CLIENTHELLO(M) ((dtls_client_hello_t *)((M) + HS_HDR_LENGTH))
 
-/* The length check here should work because dtls_*_to_int() works on
- * unsigned char. Otherwise, broken messages could cause severe
- * trouble. Note that this macro jumps out of the current program flow
- * when the message is too short. Beware!
+/*
+ * Skip variable length field.
+ *
+ * A variable length field is encoded with a preceding length followed by
+ * the value. That length itself is encoded in one to three bytes using uint8,
+ * uint16, or uint24. Decoding a variable length field requires to check first,
+ * if the length itself is within the bounds, and if so, if the value is also
+ * within the bounds.
+ *
+ * The macro "returns" the calling context with an error when the bounds are
+ * violated.
+ *
+ * \param P  pointer to length of the var field. Will be forwarded the end of
+ *           the var field.
+ * \param L  left overall data of P. Will be reduced by the size of the var
+ *           field.
+ * \param T  length type. e.g. uint8 or uint16
+ * \param A  alert description in case of a length violation
+ * \param M  logging message in case of a length violation
  */
-#define SKIP_VAR_FIELD(P,L,T) {						\
-    if (L < dtls_ ## T ## _to_int(P) + sizeof(T))			\
-      goto error;							\
-    L -= dtls_ ## T ## _to_int(P) + sizeof(T);				\
-    P += dtls_ ## T ## _to_int(P) + sizeof(T);				\
+#define SKIP_VAR_FIELD(P, L, T, A, M) {                \
+    size_t skip_length = sizeof(T);                    \
+    if (L < skip_length) {                             \
+      dtls_info("%s: field length exceeds buffer", M); \
+      return dtls_alert_fatal_create(A);               \
+    }                                                  \
+    skip_length += dtls_ ## T ## _to_int(P);           \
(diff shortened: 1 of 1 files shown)


dtls.c
```diff
@@ -3478,39 +3506,37 @@ check_certificate_request(dtls_context_t *ctx,
...
+                "CertificateRequest, certificate_types");
 
   auth_alg = 0;
   for (; i > 0 ; i -= sizeof(uint8)) {
     if (dtls_uint8_to_int(data) == TLS_CLIENT_CERTIFICATE_TYPE_ECDSA_SIGN
-	&& auth_alg == 0)
-      auth_alg = dtls_uint8_to_int(data);
+        && auth_alg == 0) {
+      auth_alg = TLS_CLIENT_CERTIFICATE_TYPE_ECDSA_SIGN;
+      data += i;
```
obgm, line 3542: This looks strange. We are trying to read single bytes here, correct (i.e., `sizeof(uint8_t)`)? When `i` happens to be greater than 1, we would advance `data` by i, i-1, i-2, ...

I suggest to re-arrange this loop as follows:
```c
for (; (i > 0) && (auth_alg == 0); i -= sizeof(uint8)) {
  if (dtls_uint8_to_int(data) == TLS_CLIENT_CERTIFICATE_TYPE_ECDSA_SIGN) {
    auth_alg = TLS_CLIENT_CERTIFICATE_TYPE_ECDSA_SIGN;
  }
  data += sizeof(uint8);
  data_length -= sizeof(uint8);
}
```

dtls.c
```diff
@@ -3531,10 +3557,13 @@ check_certificate_request(dtls_context_t *ctx,
     current_sig_alg = dtls_uint8_to_int(data);
     data += sizeof(uint8);
 
+    data_length += sizeof(uint16);
```
obgm, line 3577: I presume this needs to be `data_length -= ...`.

dtls.c
```diff
@@ -190,6 +190,34 @@ memarray_t dtlscontext_storage;
     P += dtls_ ## T ## _to_int(P) + sizeof(T);				\
   }
 
+/* The length check here should work because dtls_*_to_int() works on
+ * unsigned char. Otherwise, broken messages could cause severe trouble.
+ * Note that this macro returns an error when the message is too short.
+ *
+ * \param R  result, variable to assign the length of the field.
```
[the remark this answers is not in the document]
boaks: > I think that R gets assigned the contents of the field P points to, not its length (cf. line 212). 

That may depend more on the used terms and their definition.
In my definition of `field` this not a term for internal structs, this is used for a "logical value".
In the case of a "variable length field", that `field` is split into it's `length` and `value`, but I don't use the term `field`for this parts. What this macro does is reading that `length` of the `field` into R and it also forwards the point P to the `value` of that `field`.  I applies also all length checks in order to protect from "out of bounds" errors.

> At least, this comment is misleading, or the macro should be called something like GET_VAR_FIELD_LENGTH as it is used to get the length value of a variable-length TLS header field.

Though it does more, that name isn't really a that better match.
For me, it prepares all, including checks and failures returns, in order to the straight forward process the value of the `field`. 

I also consider to change the code even more and put that in a function. But then all calls would need to check the return and conditionally return on failures. So I decided to go for this macro, though it seems for me to be easier to read the resulting code.

dtls.c
boaks: > correct (i.e., sizeof(uint8_t))?

The `if` breaks the loop, because the proper value is found.
`i` contains the (left) field length, and in the case of the break, that left bytes are skipped.

boaks:
```
    if (dtls_uint8_to_int(data) == TLS_CLIENT_CERTIFICATE_TYPE_ECDSA_SIGN) {
      auth_alg = TLS_CLIENT_CERTIFICATE_TYPE_ECDSA_SIGN;
      // skip the rest of the field value
      data += i;
      data_length -=i;
      break;
    }
```

dtls.c
boaks: Yes.

dtls.c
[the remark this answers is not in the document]
obgm: > In my definition of `field` this is not a term for internal structs, this is used for a "logical value". In the case of a "variable length field", that `field` is split into its `length` and `value`, but I don't use the term `field`for this parts. What this macro does is reading that `length` of the `field` into R and it also forwards the point P to the `value` of that `field`. I applies also all length checks in order to protect from "out of bounds" errors.

Yes, but it would be good to put this explanation in the comment for this macro as it helps understanding what the macro does.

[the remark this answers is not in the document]
boaks: OK.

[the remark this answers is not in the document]
boaks: The comment has been edited.

obgm (approved):
LGTM

PR #198, opened 2023-04-30
Cleanup cookie

boaks:
As preview for the cleanup using SKIP_VAR_FIELDS and GET_VAR_FIELDS.


dtls.c
```diff
@@ -446,11 +447,11 @@ dtls_get_cookie(uint8 *msg, size_t msglen, uint8 **cookie) {
...
 
-  if (msglen < (*msg & 0xff) + sizeof(uint8))
-    return dtls_alert_fatal_create(DTLS_ALERT_HANDSHAKE_FAILURE);
+  GET_VAR_FIELD(cookie_len, msg, msglen, uint8, DTLS_ALERT_HANDSHAKE_FAILURE,
+                "get_cookie, cookie");
 
-  *cookie = msg + sizeof(uint8);
-  return dtls_uint8_to_int(msg);
+  *cookie = msg;
+  return cookie_len;
```
obgm, line 453: Odd: I would have expected a warning because of the cast from `size_t` to `int`. Maybe the compiler is smarter than I gave it credit for.

Otherwise: LGTM

PR #200, opened 2023-07-11
ccm.c: Remove documentation erroneous return statement for void function

mrdeep1:
Picked up by a Kali-Linux based Debian build.


boaks:
LGTM

PR #201, opened 2023-07-12
session.h: Move inclusion of string.h into .c files

obgm:
As a followup to PR #194 this change moves inclusion of the header file `string.h` into the c-files where it is required (for the declaration of `memcmp()`).

Change-Id: I11db1c37a7cc13d8fa2826f6d3e38d3765f4a4ec


mrdeep1:
LGTM

boaks:
LGTM

PR #202, opened 2023-07-22
Replace pre-processor symbols for gnu source with posix in examples

obgm:
As discussed in PR #194 the pre-processor symbol `_GNU_SOURCE` used in the examples` Makefile could be replaced by `_POSIX_C_SOURCE=200112L` to provide the functions `kill()` and `getaddrinfo()`.
The drawback is that `NI_MAXSERV` then needs to be defined manually.


mrdeep1:
LGTM

boaks:
LGTM

PR #204, opened 2023-07-27
dtls.c: fix DTLS_CH_LENGTH_MAX for renegotiation info SCSV

obgm:
To provide sufficient buffer space in dtls_send_client_hello(), the size of the TLS_EMPTY_RENEGOTIATION_INFO_SCSV pseudo cipher suite must be added to DTLS_CH_LENGTH_MAX, resulting in 12 bytes overhead instead of 10.

I think this buffer length adaptation was missing in PR #181.

Change-Id: I089a4bcac4772260fc19ea230d7521dec549e7c1

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -130,9 +130,10 @@ memarray_t dtlscontext_storage;
  * cookie                 := n bytes
  * cipher_length          := 2 bytes
  * cipher suites (max)    := 2 bytes + max * 2
+ * empty_renegotiation    := 2 bytes
  * compression_length     := 1 byte
  * compression            := 1 byte
- * extensions_length      := 2 bytes   => 10 + max * 2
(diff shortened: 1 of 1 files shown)


dtls.c
```diff
@@ -130,9 +130,10 @@ memarray_t dtlscontext_storage;
  * cookie                 := n bytes
  * cipher_length          := 2 bytes
  * cipher suites (max)    := 2 bytes + max * 2
+ * empty_renegotiation    := 2 bytes
```
boaks, line 133: I think, the `2 bytes` in `2 bytes + max * 2` are already the new `empty_renegotiation := 2 bytes`.

I guess, this is a rudiment of changing the same code a couple of times over a longer period and also change the order of processing those changes.

obgm: Oh, ok, I will close this PR.

PR #215, opened 2023-12-07
Use __USE_MINGW_ANSI_STDIO on windows-

boaks:
Support printf format specifier "%zu" for size_t.


boaks:
Move the flag into the cmakefile.
See PR #216

PR #216, opened 2023-12-07
Enable __USE_MINGW_ANSI_STDIO in CMakefile for MinGW.

boaks:


boaks:
Withdrawn, using PR #218.

PR #222, opened 2024-01-18
Apply option local port

boaks:
The previous version ignores the option for the local port. That may be
caused by issues using the same default local port for the server and
client.
This enables the use of an specific local port and changes the default
to an ephemeral free port, similar to quite a lot of other UDP clients.
The DEFAULT_PORT is therefore only used for the destination.

### DIFF
--- a/tests/dtls-client.c
+++ b/tests/dtls-client.c
@@ -376,9 +376,11 @@ usage( const char *program, const char *version) {
 #endif /* DTLS_PSK */
           "\t-o file\t\toutput received data to this file\n"
           "\t       \t\t(use '-' for STDOUT)\n"
-          "\t-p port\t\tlisten on specified port (default is %d)\n"
+          "\t-p port\t\tlisten on specified port\n"
+          "\t       \t\t(default is an ephemeral free port).\n"
           "\t-r\t\tforce renegotiation info (RFC5746)\n"
-          "\t-v num\t\tverbosity level (default: 3)\n",
+          "\t-v num\t\tverbosity level (default: 3)\n"
+          "\tDefault destination port: %d\n",
           DEFAULT_PORT);
 }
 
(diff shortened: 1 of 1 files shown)


tests/dtls-client.c
```diff
@@ -473,40 +478,53 @@ main(int argc, char **argv) {
...
       force_renegotiation_info = 1;
       break;
     case 'v' :
       log_level = strtol(optarg, NULL, 10);
       break;
+    case -1 :
+      /* handle arguments */
+      if (!dst.size) {
+        /* resolve destination address where server should be sent */
+        res = resolve_address(argv[optind++], &dst.addr.sa);
```
mrdeep1, line 493: I'm not convinced that this is the best way do do this (very unusual), moving the logic processing the arguments to within the `getopt()` logic, and fail to see why it is actually needed.  If this is done, at a minimum `dtls_set_log_level(log_level);` needs to be included at the start.

I think that simply removing the use of `port_str`, using the new `local_port` or `DEFAULT_PORT `in the appropriate places along with the other changes should should work

boaks: I guess, the comment is for PR #221.

This PR and PR #221 are changing the very same lines so I decided to put the changes of this PR on the top of PR #221.

mrdeep1: Ah - yes, this really should be against #221.

boaks: I already copied your comment ;-)

obgm (approved):
LGTM.

Just a small comment: Technically, there is no "IPv4 port" or "IPv6 port" as described in the comments prior to the bind. As I think that this is useful to differentiate these two cases, I am entirely ok with leaving it like this.

boaks:
I guess, adding a ',' will be an improvement :-)

```
dtls_info("bind to local IPv6, port %u\n", local_port);
```

PR #223, opened 2024-02-06
dtls.c: Handle DTLS1.3 ClientHello when calculating cookie

mrdeep1:
Do not calculate the cookie using the Extensions as these are different between DTLS1.2 and DTLS1.3

https://datatracker.ietf.org/doc/html/rfc6347#section-4.2.1

When responding to a HelloVerifyRequest, the client MUST use the same parameter values (version, random, session_id, cipher_suites, compression_method) as it did in the original ClientHello.  The server SHOULD use those values to generate its cookie and verify that they are correct upon cookie receipt.

https://www.rfc-editor.org/rfc/rfc9147.html#section-5.3

The ClientHello up to, but not including the Extensions is the same for DTLS1.2 and DTLS1.3

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -464,7 +464,7 @@ dtls_create_cookie(dtls_context_t *ctx,
 		   uint8 *msg, size_t msglen,
 		   uint8 *cookie, int *clen) {
   unsigned char buf[DTLS_HMAC_MAX];
-  size_t e, fragment_length;
+  size_t e, compatability_length;
   int len;
 
   /* create cookie with HMAC-SHA256 over:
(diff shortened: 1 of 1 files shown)


boaks:
That mainly fails then later in the handshake, or?

boaks:
I'm still wondering/confused:
I understand your description of the processing of DTLS 1.2 /  1.3 mixed handshakes.
But that would mean "DTLS 1.3 breaks the backwards compatibility to 1.2". At least if the 1.3 is the client and 1.2 is the server. Not sure, if that is intended by the TLS working group.

boaks:
I guess, there are two understandings of

"same parameter values (version, random, session_id, cipher_suites, compression_method)"

one is, that the list defines exactly the " parameter values", the other, that it defines an example and the extensions may be added. I guess, the first one is the common one. I wrote a e-mail to one of the RFC9147 authors. Let's see if I get an answer. 
If not, LGTM.

mrdeep1:
While unclear what should happen with a HelloVerifyRequest (as opposed to a HelloRetryRequest), https://www.ietf.org/archive/id/draft-ietf-tls-esni-17.html#section-6.1.5  will update the ClientHello for the second ClientHello request

dtls.c
```diff
@@ -504,14 +504,23 @@ dtls_create_cookie(dtls_context_t *ctx,
...
+   * Read in compatablility_length length and check for consistency
+   * The compatability_length is because a DTLS1.3 Client Hello may come in,
+   * and the re-transmit of the Client Hello with cookie may be different.
+   */
+   /* Get the Cipher Suites length + value */
+   compatability_length = dtls_uint16_to_int(msg + DTLS_HS_LENGTH + e) + 2;
+   /* Add in the Compression length byte + value */
+   compatability_length += dtls_uint8_to_int(msg + DTLS_HS_LENGTH + e
+                                             + compatability_length) + 1;
+
```
boaks, line 517: The other places, where the length of the field is read, do a length check before reading it.

```
  if (e + DTLS_HS_LENGTH + sizeof(uint8_t) > msglen)
    return dtls_alert_fatal_create(DTLS_ALERT_HANDSHAKE_FAILURE);
  /* skip cookie bytes and length byte */
  e += dtls_uint8_to_int(msg + DTLS_HS_LENGTH + e);
```

In order to "standardize" that, we have two macros, SKIP_VAR_FIELD and GET_VAR_FIELD. In this case I guess, using SKIP_VAR_FIELD (together with other modifications) will do it.

boaks: See PR #230

boaks:
Merged alternative implementation with PR #230.

PR #228, opened 2024-02-18
Support for micro-ecc (WIP)

obgm:
This is a proof of concept for supporting [micro-ecc](https://github.com/kmackay/micro-ecc) as a replacement for the ECC implementation in tinydtls.
micro-ecc is included as a submodule, therefore you either need to specify `--recurse-submodules` with `git clone` or do a `git submodule update --init --recursive` in an existing working copy.

For backwards compatibility, only curve secp256r1 is supported.

This change is based on PR #227 to avoid warnings during the build process.


crypto.c
```diff
@@ -517,20 +608,39 @@ dtls_ecdsa_verify_sig_hash(const unsigned char *pub_key_x,
...
+  }
   (void)result_r;
   (void)result_s;
 
-  assert(key_size == DTLS_EC_KEY_SIZE);
-  assert(sign_hash_size >= uECC_BYTES);
-  assert(sizeof(sign) >= 2 * uECC_BYTES);
+  curve_size = uECC_curve_public_key_size(uecc_curve);
+
+  assert(key_size == (unsigned int)curve_size);
```
boaks, line 638: That fires on my interop test. Adding 

```
dtls_warn("%d != %d\n", (int) key_size, curve_size);
```

I get 

```
Feb 26 16:46:45 WARN 32 != 64
tinydtls-server: crypto.c:638: dtls_ecdsa_verify_sig_hash2: Assertion `key_size == (unsigned int)curve_size' failed.
```

AFAIK, the `key-size` is passed in as `sizeof(config->keyx.ecdsa.other_pub_x)` in dtls.c, line 2589

boaks: Beside the assert, the first interop-tests with Californium are successful.
Great!

crypto.c
```diff
@@ -24,14 +24,17 @@
...
 
 #include "global.h"
 #include "dtls_debug.h"
 #include "numeric.h"
 #include "dtls.h"
 #include "crypto.h"
 #include "ccm.h"
-#include "ecc/micro-ecc/uECC.h"
+#ifdef DTLS_ECC
+#include "ext/micro-ecc/uECC.h"
```
boaks, line 36: "ext/micro-ecc" is part of the include path, maybe it's simpler to use "uECC.h"?

mrdeep1: It is my preference to always include the package information path in the #include statement, to prevent name clashes (unlikely with `uECC.h`, but with possible with names like `types.h` etc.).

So, I would recommend that the include path includes `ext/`, and use `#include <micro-ecc/uECC.h>`.

boaks: The pain is, that "micro-ecc/uECC.c" uses: 

```
#include "uECC.h"
#include "uECC_vli.h"
```

and "micro-ecc/uECC_vli.h" uses:

```
#include "uECC.h"
#include "types.h"
```

mrdeep1: Anything that is getting built external to micro-ecc will need the equivalent of "micro-ecc/uECC.h" to get at the function definitions.  For the actual micro-ecc code compilation, I assume that it is done in the micro-ecc/ directory, in which case just "uECCH.h" without a path is fine.
````
If you declare #include "" , the compiler first searches in current directory of source file and if not found,
continues to search in any included -I directories and then the default include directories.

If you declare #include <> , the compiler searches directly in any included -I directories and then the default
include directories.

$ cpp -v shows the include search directory order.
````

Comment (boaks) [reply crypto.c:36]: OK. I've checked it and at least the cmake build doesn't need to extend the include path with "ext/micro-ecc".

Comment (boaks) [reply crypto.c:36]: > "ext/micro-ecc" is part of the include path,

I mixed it up, it's only part of "FILES".

File: dtls.c [hunk 3]
```diff
@@ -337,6 +349,9 @@ dtls_init(void) {
 memarray_init(&dtlscontext_storage, dtlscontext_storage_data,
               sizeof(dtls_context_t), DTLS_CONTEXT_MAX);
 #endif /* RIOT_VERSION */
+#ifdef DTLS_ECC
+ uECC_set_rng(uecc_rng_function);
+#endif /* DTLS_ECC */
```
boaks, line 354: Maybe that could be moved to "crypto.c". In my opinion, that would limit the scope of "ecc-related changes" to crypto.

PR #230, opened 2024-02-28
dtls.c: exclude Extensions when calculating cookie.

boaks:
Do not calculate the cookie using the Extensions as these are different between DTLS1.2 and DTLS1.3

https://datatracker.ietf.org/doc/html/rfc6347#section-4.2.1

When responding to a HelloVerifyRequest, the client MUST use the same parameter values (version, random, session_id, cipher_suites, compression_method) as it did in the original ClientHello.  The server SHOULD use those values to generate its cookie and verify that they are correct upon cookie receipt.

https://www.rfc-editor.org/rfc/rfc9147.html#section-5.3

The ClientHello up to, but not including the Extensions is the same for DTLS1.2 and DTLS1.3

Alternative implementation for PR #223 with boundary checks using SKIP_VAR_FIELD.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -464,7 +464,7 @@ dtls_create_cookie(dtls_context_t *ctx,
 		   uint8 *msg, size_t msglen,
 		   uint8 *cookie, int *clen) {
   unsigned char buf[DTLS_HMAC_MAX];
-  size_t e, fragment_length;
+  uint8 *start;
   int len;
 
   /* create cookie with HMAC-SHA256 over:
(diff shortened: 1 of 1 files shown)


mrdeep1:
This version works for me instead of #223.

boaks:
@obgm 

Any opinion?

mrdeep1:
[pre-post.zip](https://github.com/eclipse/tinydtls/files/15249781/pre-post.zip)

pcap traces before / after the fix

PR #234, opened 2024-03-04
README.md: add some details and references.

boaks:
Using tables doesn't allow to limit the line length.
If the line length is considered to be MUST, then I guess I need to remove the tables and use lists.


obgm:
LGTM

Issue #224, opened 2024-02-08
Experiencing an extended handshake duration on a SAMR21-xpro device.


LizzieDE110:
Hello, I am currently working on a project to secure wireless UDP/IP communication between SAMR21-xpro devices using RIOT OS. After some research, I found that using DTLS (with ECDSA keys) was the best solution. Moreover, RIOT has examples that use tinyDTLS.

My problem is that, with SAMR21-xpro, the handshake is extremely long. From the logs and packet analysis, I feel like the problem lies in the key management (generation, signature, and verification). According to tinydtls/ecc/ecc.h, this implementation is supposed to be very efficient for 32-bit CPUs.

I made a quick example on RIOT that just generates a key pair and signs a hashed data. The generation takes ~6 seconds, and the signature also takes 6 seconds. I haven't tested verification, but I assume it also takes many seconds. I am quite confused because initially, when I was trying to run a sock-dtls example from RIOT, I found it would be better to generate a key pair to use, and to do so, I used the micro-ecc library. On SAMR21-xpro, key generation, signature, and verification take barely a second.

Is it normal that the key management is so long on a 32-bit CPU?

Also, I was wondering if it could be a good idea to replace the library used to manage ECC in tinyDTLS with micro-ecc? Or maybe add a patch that could let the user choose what they want to use?

Thanks in advance

boaks:
From the datasheets of the SAMR21 I found, that it's a Cortex M0+ with 48MHz. Several seconds on a Cortex-M33 (without using the ECC HW accelerator, for now not supported) is also my experience. Depending on quite a lot of other topics, using special HW as a ATECC608 would help, but it's also for now not support.

> Also, I was wondering if it could be a good idea to replace the library used to manage ECC in tinyDTLS with micro-ecc? 

Do you have some benchmark numbers?

LizzieDE110:
What types of benchmark numbers are you interested in?
(I'm relatively new to programming, so I'm not very familiar with these processes.)

boaks:
Pretty simple numbers.

> The generation takes ~6 seconds, and the signature also takes 6 seconds. I haven't tested verification, but I assume it also takes many seconds. 

Just some number as that. My point is, if it is really faster, e.g. 3s vs. 6s it may be worth.
But to be frank, I consider to spend rather time in the ATECC608, because that brings also a "secure store" for key and a real performance boost.

LizzieDE110:
Example using ecc from tinydtls/ecc: 
[...]

PR #236, opened 2024-03-04
WIP: Update crypto API to use a Microship secure element such as ATECC608A

LizzieDE110:
I am working on boards that embed an ATECC608A (Microchip secure element), and I would like to perform all ECC/ECDH operations using it. This approach offers the benefit of enhanced security, as private keys are not directly manipulated.

Initially, I am working with RIOT. When I started looking for solutions to secure RF communication, I came across tinydtls and was convinced by its implementation of DTLS. I noticed some issues on my board (reported here [Issue 224](https://github.com/eclipse/tinydtls/issues/224)), and a PR to replace ecc with micro-ecc is now open ([PR 229](https://github.com/eclipse/tinydtls/pull/229)).

Micro-ecc works on my boards and significantly accelerates the handshake process. However, as mentioned earlier, I have an ATECC608A on my boards, so I would like to add a feature allowing users to choose whether to use it or not.

Since not all boards embed a secure element, I have added an environment variable (DTLS_ATECC608A) to give users the choice to use it or not. The CryptoAuth library is used to control the secure element ([CryptoAuthLib](https://github.com/MicrochipTech/cryptoauthlib)).

This PR is still a work in progress, but I have opened it so that we can discuss and improve my work.

### DIFF
--- a/crypto.h
+++ b/crypto.h
@@ -30,6 +30,10 @@
 #include "hmac.h"
 #include "ccm.h"
 
+#ifdef DTLS_ATECC608A
+#include "cryptoauthlib.h"
+#endif /* ATECC608A */
+
 /* TLS_PSK_WITH_AES_128_CCM_8 */
 #define DTLS_MAC_KEY_LENGTH    0
 #define DTLS_KEY_LENGTH        16 /* AES-128 */
@@ -219,6 +223,28 @@ typedef struct {
 
 /* just for consistency */
 #define dtls_kb_digest_size(Param, Role) DTLS_MAC_LENGTH
+
+#ifdef DTLS_ATECC608A
+/**
+ * @brief Slot id used to perform ECDHE operation. 
+ *        Due to the last 'E', the operation is performed with 
+ *        ephemeral keys that must be generated and stored in the
+ *        the current slot.
+ *        I advise to use slot configured as follows:
+ *          - SlotConfig[slot_id] = 0x2087
+ *          - KeyConfig[slot_id] = 0x0013
+ * @warning Slot ID must be different from ecc_slot_id.
+ */
+static uint8_t ecdhe_slot_id;
+
+/**
+ * @brief Slot id used to perform ECDSA operation.
+ *        This slot must contains the private key used to sign the
+ *        message. The associated public key is used to verify the signature.
+ * @warning Slot ID must be different from ecdhe_slot_id.
+ */
+static uint8_t ecc_slot_id;
+#endif
 
 /** 
  * Expands the secret and key to a block of DTLS_HMAC_MAX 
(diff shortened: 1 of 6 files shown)


crypto.h
```diff
@@ -220,6 +224,32 @@ typedef struct {
...
+/**
+ * @brief Slot id used to perform ECDSA operation.
+ *        This slot must contains the private key used to sign the
+ *        message. The associated public key is used to verify the signature.
+ *        When using test_ecc608_configdata of cryptoauthlib tests, 
+ *        you can use slot 7.
+ * @warning Slot ID must be different from ecdhe_slot_id.
+ */
+extern uint8_t ecc_slot_id;
+#endif
```
boaks, line 251: As short term work around, maybe.
But I guess, if a peer executes 2 handshakes simultaneously, that will fail.

LizzieDE110: CryptoAuthLib offers a mutex implementation. However, I suspect that if the handshake takes too long, a connection might fail due to timeout?

LizzieDE110:
1. I believe encoding the slot ID shouldn't be too difficult. I can give it a try.

2. I'm not entirely sure I understand your point. As far as I know, the keys are stored in the ATECC608A. I'm not convinced we need to free those keys. If we need to generate new keys, we can simply call the atcab_genkey function, which will overwrite the previous keys (assuming the slot is configured to allow it). The main advantage of the ATECC608A is that we don't need to store keys in our processor's RAM, especially with private keys that we can't read at all.

That sounds reasonable. However, my primary concern is that the ATECC608A can be fully personalized depending on the chosen configuration. With my proof-of-concept configuration, I only have two slots capable of performing a 'correct' handshake. Furthermore, because I locked the configuration zone to test my implementation, I can't change the device's configuration without obtaining a new one. Nevertheless, I can add accessors so that users can provide a list of slots rather than just one slot?

boaks:
1. cool

2. there are two type of keys: 
- the long term keys for ECDSA, they must stay in the ATECC608A.
- the ephemeral keys for ECDHE. This keys are generated at the begin of the handshake and then used once later in the handshake to generate the premaster secret. AFAIK, the ATECC608A has only a limited number of key slots. If, e.g. the implementation reserves 10 slots for ECDHE, only 10 handshakes in parallel are possible. That's also great for a client, because it's up to the client to initiate handshakes and not to overload the ATECC608A. 
For a server this will require to decide, what happens, if more than 10 clients want to do a handshake in parallel. A common idea will then be, to reject or ignore more handshakes. But that opens then a door for so called "exhaust attacks". Someone may start handshakes from several endpoints just to trigger the server to create ECDHE keys and block the slots. Once the server create the keys and sent the server key exchange, the attacker ignores further message, which extends then the time period, where the slot is blocked. Real clients will be faced to be rejected, because no free key slot is available.
Anyway, this is just a consideration when used as server. I consider the ATECC608A be mainly used on the client side.

boaks:
About 3. 
I'm still not common enough to the ATECC608A. I would expect, that it provides the function to protect some key slots for ECDSA (some to identify the device, some for trusts to validate the other peers). It depends now on the intended usage. For a peer, which is intended to support a specific number of parallel handshakes, it requires to have free slots for the generated ECDHE keys. Even if that number maybe 1.

Fl-Al-Ve:
This PR seems interesting.
Are there any updates on this topic?

PR #239, opened 2024-04-04
dtls.c, crypto.h: add documentation for handling dtls_user_parameters_t.

boaks:


obgm (commented):
There is a small typo in both comments. Depending on whether the users are one or many, it must read "user's" or "users'". (Genitive in both cases.)

Otherwise, LGTM.

boaks:
Does this fix the type?

Issue #186, opened 2023-01-25
RFC7250 - RawPublicKey - No Suitable Certificate?


boaks:
I'm currently on testing the interoperability of the [OpenSSL - RPK PR](https://github.com/openssl/openssl/pull/18185).

If the client has "no suitable certificate", the current implementation in "main" branch aborts the handshake.

[dtls.c: check_server_hellodone](https://github.com/eclipse/tinydtls/blob/main/dtls.c#L3356)

An alternative approach would be, to implement [RFC 5246 - 7.4.6 - Client Certificate](https://www.rfc-editor.org/rfc/rfc5246#section-7.4.6) :

> If no suitable certificate is available,  the client MUST send a certificate message containing no certificates.  That is, the certificate_list structure has a length of zero.

That enables an server to decide to continue or abort.

Any opinions?

mrdeep1:
Empty client certificate list back to the server works for me.

PR #241, opened 2024-08-26
dtls.c: support empty client certificate.

boaks:
Use empty client certificate, if a certificate is requested but no certificate is available.

Issue #186


obgm:
LGTM

PR #242, opened 2024-08-26
dtls.c: ensure, alerts are reported by events.

boaks:
For warning/close_notify wait for the
retransmission time or response before
closing the peer.

For other alerts report event immediately
and close the peer.


obgm:
LGTM

PR #262, opened 2025-10-06
Fix MSVC compilation (CMake-based)

raspopov:
Using the Ninja generator with the MSVC compiler leads to errors due to incorrect compiler call options. In CMake, the generator and the compiler are different entities.

This fix replaces the incorrect CMake variable with the correct one that describes the current MSVC-compatible compiler.


raspopov:
I'm sure I signed it:
```
The Eclipse Contributor Agreement 3.1 that you signed is valid for a period of three years from date of signing and is set to expire on 2028-10-06.
```
I can confirm that my email address and my GitHub linked account are both correct. 😔

PR #263, opened 2025-10-07
Fix CMake test building

raspopov:
A CMake unit test build was added.

The original tests (examples) were fixed so they can be compiled on Windows using MSVC2022 and MinGW. The `ccm-test` has been updated so that it returns the status of test completion.The file `getopt.c` was borrowed from the libcoap project.

Unimportant compiler warnings were disabled.

Tests (both) can now be run in the standard way via CTest.


mrdeep1:
> Am I correct in interpreting these CI errors as meaning that the libcunit1-dev package was not installed in the build environment?

Yes.  They are installed for the build-linux-autotools builds.

PR #265, opened 2025-11-20
doc/Makefile.in: Update Doxyfile to latest version before use

mrdeep1:
Add in '$(DOXYGEN) -u' to update Doxyfile to the latest version to reduce any old configuration warning errors.

### DIFF
--- a/doc/Makefile.in
+++ b/doc/Makefile.in
@@ -18,7 +18,8 @@ DISTDIR?=$(top_builddir)/@PACKAGE_TARNAME@-@PACKAGE_VERSION@
 FILES:=Makefile.in Doxyfile.in html
 
 doc:	Doxyfile
-	$(DOXYGEN) -l
+	@$(DOXYGEN) -u $< > /dev/null 2>&1
+	@$(DOXYGEN) -l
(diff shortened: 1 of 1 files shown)


boaks:
@obgm 

Any thoughts on this?

mrdeep1:
It does not change any TinyDTLS code, but gets rid of a lot of build warnings.

boaks:
I'm just not common with doxygen, so I'm not sure, if that causes also other issues.
What will happen, if the build is done in a not "connected environment"?

mrdeep1:
All that happens is the current doxygen executable updates the doxygen configuration file in situ and has no need to reference any external information.

If there is any issue, this additional line would fail and the build would continue as if the line was not there.

obgm:
I am not a big fan of this type of dynamic changes to the `Doxyfile` (that was auto-generated from `Doxyfile.in`).  I would prefer to manually update the source, `Doxyfile.in`.

However, `doxygen -u` did never break anything for me, so running it at that point of the `Makefile` seems not that critical, and therefore, I do not object to applying this change.

obgm (approved):
LGTM

PR #267, opened 2026-03-17
dtls.c: fix length check in check_server_certificate.

boaks:
Fixes potential out-of-bounds read.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -3430,6 +3430,11 @@ check_server_certificate(dtls_context_t *ctx,
   int err;
   dtls_handshake_parameters_t *config = peer->handshake_params;
 
+  if (data_length < DTLS_HS_LENGTH + DTLS_EC_SUBJECTPUBLICKEY_SIZE + sizeof(uint24)) {
+    dtls_alert("the packet length does not match the expected\n");
+    return dtls_alert_fatal_create(DTLS_ALERT_DECODE_ERROR);
+  }
+
   update_hs_hash(peer, data, data_length);
 
   assert(is_key_exchange_ecdhe_ecdsa(config->cipher_index));


boaks:
@obgm 

Please a fast LGTM, thanks!

dtls.c
```diff
@@ -3436,6 +3436,11 @@ check_server_certificate(dtls_context_t *ctx,
 
   data += DTLS_HS_LENGTH;
 
+  if (data_length < DTLS_HS_LENGTH + DTLS_EC_SUBJECTPUBLICKEY_SIZE + sizeof(uint24)) {
```
obgm: Looks good in general but I think this check should come before doing the `update_hs_hash` in [line 3433](https://github.com/eclipse-tinydtls/tinydtls/pull/267/changes/612e9f7c679dc9c132159945014e84d706ce5f68#diff-947f252e5a6242133f45e2963433822253886eb9a6bf4c12c7163ac8e90fe017L3433) . 

Otherwise, LGTM.

boaks: Done.

wsparks-vc:
Hello @boaks and @obgm - does the Eclipse Foundation CNA intend to assign a CVE for the reported vulnerability?

boaks:
From the team's (obgm an me) view, tinydtls is in development stage. We have no 1.0 release. 
I personally don't know, if we find enough "volunteer time", at least I'm not longer able to spend that larger amount of time.
Therefor we also decided two years ago, just to fix vulnerabilities (if possible with the time we#re able to spend), but don't go for CVE reporting. It's just really a question of work-time.

wsparks-vc:
It's standard practice in vuln management to issue CVEs to public patches, rather than waiting for an official release. Through managing the VulnCheck CVE Numbering Authoring (CNA), I usually work with maintainers to allow for official version releases to be put together prior to CVE publication. Since the Eclipse Foundation would, presumably, be the CNA with appropriate first-level scope, we'd like to push for CVE issuance ASAP.

PR #271, opened 2026-05-28
fix(dtls): remove duplicate DEL_PEER in handle_alert()

### LINKED ISSUE #269: Double DEL_PEER in handle_alert()
File : dtls.c
Function : handle_alert() (line ~4517) and dtls_destroy_peer() (line ~2205)
When a DTLS peer sends a fatal alert or close_notify, the handle_alert() function removes the peer from the hash table via DEL_PEER() , then calls dtls_destroy_peer() which attempts to remove the same peer again . This double removal corrupts the uthash hash table and can lead to crashes or undefined behavior.

The DEL_PEER macro expands to HASH_DELETE() , which removes an entry from the uthash hash table. When called twice on the same peer:
1. First call : Correctly removes the peer from the hash table
2. Second call : Operates on an entry that is no longer in the hash table
The HASH_DELETE_HH implementation in uthash.h does not check if the entry is actually in the table before attempting removal. This leads to:
 Corrupted hash table bucket chains
 Incorrect num_items counter
 Potential use-after-free if memory is reallocated between calls
 Possible NULL pointer dereference in edge cases

Please confirm whether this is a bug or an intentional design that I haven't understood.

re2zero:
## Summary
- Remove duplicate `DEL_PEER()` call in `handle_alert()` to fix uthash corruption

## Problem
`handle_alert()` calls `DEL_PEER()` to remove the peer from the hash table, then calls `dtls_destroy_peer()` which calls `DEL_PEER()` again on the same peer. Since `HASH_DELETE_HH` in uthash does not check if the entry is in the table, this double removal corrupts the hash table bucket chains and `num_items` counter, potentially leading to crashes or use-after-free.

## Fix
Remove the `DEL_PEER()` call from `handle_alert()` since `dtls_destroy_peer()` already handles the complete peer cleanup including removal from the hash table.

## Changes
| File | Change |
|------|--------|
| `dtls.c` | Remove duplicate `DEL_PEER()` in `handle_alert()`, update comment |

Fixes #269

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -4502,10 +4502,9 @@ handle_alert(dtls_context_t *ctx, dtls_peer_t *peer,
   dtls_info("** Alert: level %d, description %d\n", data[0], data[1]);
 
   /* The peer object is invalidated for FATAL alerts and close
-   * notifies. This is done in two steps.: First, remove the object
-   * from our list of peers. After that, the event handler callback is
-   * invoked with the still existing peer object. Finally, the storage
-   * used by peer is released.
+   * notifies. The event handler callback is invoked with the peer
+   * object, then dtls_destroy_peer() removes the peer from the
+   * hash table and releases the storage.
    */
   close_notify = data[1] == DTLS_ALERT_CLOSE_NOTIFY;
   if (data[0] == DTLS_ALERT_LEVEL_FATAL || close_notify) {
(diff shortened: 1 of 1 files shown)


re2zero:
Closing this PR as I am unable to complete the Eclipse Contributor Agreement (ECA) requirement. The fix itself is straightforward: remove the duplicate DEL_PEER() call in handle_alert() to prevent uthash corruption. Feel free to take the patch.

boaks:
PR #270 is already addressing that issue.

PR #275, opened 2026-06-27
dtls.c: add missing length checks in dtls_prepare_record.

boaks:


dtls.c
```diff
@@ -1774,6 +1774,10 @@ dtls_prepare_record(dtls_peer_t *peer, dtls_security_parameters_t *security,
    	            } CCMNonceExample;
     */
 
+    if (*rlen < DTLS_RH_LENGTH + 8) {
```
obgm, line 1777: I have difficulties wrapping my head around this length check. The storage space for the 8 bytes written after this check are already included in DTLS_RH_LENGTH. Which fields are checked here actually?

boaks: On a fast look, it seems that this is already considered in line 1680 with

```
 if (*rlen < DTLS_RH_LENGTH) {
```

but the pain comes from the use of `epoch+record-sequence-number` as explicit nonce.
Therefore I check, if there is space for the 8 bytes of the explicit nonce. These bytes are located direct after the header.

Does this help?

PR #287, opened 2026-09-14
dtls-client.c: fix endless loop on second cli argument.

boaks:


tests/dtls-client.c
```diff
@@ -439,7 +440,9 @@ main(int argc, char **argv) {
 #endif /* DTLS_PSK */
 
   while (optind < argc) {
-    opt = getopt(argc, argv, "c:eo:p:rv:z" PSK_OPTIONS);
+    if (0 <= opt) {
+      opt = getopt(argc, argv, "c:eo:p:rv:z" PSK_OPTIONS);
+    }
```
mrdeep1: My preference (for readability) would be to do
````
-    while (optind < argc) {
-      opt = getopt(argc, argv, "c:eo:p:rv:z" PSK_OPTIONS);
+   while ((opt = getopt(argc, argv, "c:eo:p:rv:z" PSK_OPTIONS)) != -1) {
````
but what is there works.

boaks: The issue with the old code was, that calling `getopt` a second time, when it already returned `-1`, the `optind` is set back to the value of the first argument. With that the `argv[optind++]` doesn't work as intended.
Maybe it's clearer to redesign the `case -1:` instead of that "minimal workaround".

Issue #278, opened 2026-08-11
`MASK_L` over-shifts for RFC 3610-legal `L ≥ 4`, collapsing the CCM counter (keystream reuse at `L = 4`)


LukeW1999:
`ccm.c:31`:

```c
#define MASK_L(_L) ((1 << 8 * _L) - 1)
```

`1` has type `int`. RFC 3610 §2.1 permits `L ∈ [2, 8]`, so `8 * _L` reaches 64
and any `L ≥ 4` shifts a 32-bit `int` by at least its width — undefined
behaviour. `ccm.h` documents `M` and `L` by referring the caller to RFC 3610 and
places no bound of its own, and `dtls_ccm_decrypt_message` has no documentation
block at all.

### This is already happening in the test suite

`tests/ccm-testdata.c` has four vectors with `L = 5`, and the loops at
`tests/unit-tests/test_ccm.c:33` and `:56` run every vector through
`dtls_ccm_encrypt_message` / `dtls_ccm_decrypt_message` unfiltered. Running that
loop under UBSan:

```
ccm.c:221:3: runtime error: shift exponent 40 is too large for 32-bit type 'int'
ccm.c:147:3: runtime error: shift exponent 40 is too large for 32-bit type 'int'
ccm.c:291:3: runtime error: shift exponent 40 is too large for 32-bit type 'int'
35 vectors x 2 directions: 70 pass, 0 fail
```

The vectors pass while executing the undefined shift, which is presumably why it
has gone unnoticed: at `L = 5` the counter is truncated to 8 bits and no vector
is long enough for that to change an output.

### The consequence at `L = 4` is keystream reuse, not just UB

`SET_COUNTER` only fills the counter field while `C` is non-zero:

```c
memset((A) + DTLS_CCM_BLOCKSIZE - (L), 0, (L));
(C) = (cnt) & MASK_L(L);
for (i_ = DTLS_CCM_BLOCKSIZE - 1; (C) && (i_ > (L)); --i_, (C) >>= 8)
  (A)[i_] |= (C) & 0xFF;
```

On x86-64 the shift count is taken mod 32, so the macro evaluates to:

| L | `MASK_L(L)` | intended |
|---|---|---|
| 2 | 65535 | 65535 |
| 3 | 16777215 | 16777215 |
| **4** | **0** | 4294967295 |
| 5 | 255 | 1099511627775 |
| 6 | 65535 | 281474976710655 |
| 7 | 16777215 | 72057594037927935 |

[...]

PR #288, opened 2026-09-18
Sanitizer: Add in build environment support

mrdeep1:
configure --enable-sanitizer
cmake -DENABLE_SANITIZER=ON

Addresses trivial issues raised by the ASan sanitizer.

Includes m4/ax_check_compile_flag.m4, which is a part of the
pkg autoconf-archive which not all systems have.


ccm.c
```diff
@@ -28,7 +28,7 @@
 
 #define CCM_FLAGS(A,M,L) (((A > 0) << 6) | (((M - 2)/2) << 3) | (L - 1))
 
-#define MASK_L(_L) ((1 << 8 * _L) - 1)
+#define MASK_L(_L) (((size_t)1 << (8 * _L)) - 1)
```
boaks: That was reported in issue #278

hmac.c
```diff
@@ -46,7 +46,7 @@ dtls_hmac_init(dtls_hmac_context_t *ctx, const unsigned char *key, size_t klen)
     dtls_hash_init(&ctx->data);
     dtls_hash_update(&ctx->data, key, klen);
     dtls_hash_finalize(ctx->pad, &ctx->data);
-  } else
+  } else if (key)
```
boaks, line 49: Not sure, but if `key` must be checked, then that must be above that, the other if branch is also using `key`.

boaks: Maybe the logic is, if klen is 0, key maybe NULL.

mrdeep1:
Just updated to include m32 testing.

hmac.c
mrdeep1: ASan complained.

ccm.c
mrdeep1: Thanks for the pointer.  It needs to be uint64_t for m32 builds.

